PR #2, opened 2016-09-26
Use `module_function`.

reitermarkus:
Needed for https://github.com/Homebrew/brew/pull/1145.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -93,6 +93,8 @@ require "development_tools"
 require "utils/bottles"
 
 module Homebrew
+  module_function
+
   BYTES_IN_1_MEGABYTE = 1024*1024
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Can you explain this a bit; not sure I understand?

MikeMcQuaid:
@reitermarkus Think this should be good to 🚢 too?

reitermarkus:
Yes. 👍

MikeMcQuaid:
@reitermarkus If you felt like doing the same thing for homebrew-bundle and homebrew-services that'd be 💟

reitermarkus:
None of them use `module Homebrew`, so should be good.

MikeMcQuaid:
@reitermarkus 🆒 thanks.

PR #4, opened 2016-09-27
Test that brew man changes no files

scpeters:
I believe that `brew man` is not part of CI. This runs it in the test-bot and verifies that it doesn't cause any changes to the tracked files. I put this very early in the test-bot loop to speed up testing, but we could put it elsewhere. I believe the master branch of brew currently will fail this test.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -710,6 +710,11 @@ module Homebrew
       return if @skip_homebrew
 
       if !@tap && (@formulae.empty? || @test_default_formula)
+        # this should always succeed
+        test "brew", "man"
+        # this succeeds if the committed manpages are up-to-date
+        test "brew", "man", "--fail-if-changed"
+
         # TODO: try to fix this on Linux at some stage.
         if OS.mac?
           # test update from origin/master to current commit.
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -710,9 +710,7 @@ module Homebrew
       return if @skip_homebrew
 
       if !@tap && (@formulae.empty? || @test_default_formula)
-        # this should always succeed
-        test "brew", "man"
-        # this succeeds if the committed manpages are up-to-date
+        # verify that manpages are up-to-date
         test "brew", "man", "--fail-if-changed"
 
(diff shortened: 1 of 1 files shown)


cmd/brew-test-bot.rb
```diff
@@ -710,6 +710,11 @@ def homebrew
       return if @skip_homebrew
 
       if !@tap && (@formulae.empty? || @test_default_formula)
+        test "brew", "man"
+        test "git", "diff", "--exit-code", "-C", HOMEBREW_REPOSITORY, "docs/brew.1.html"
+        test "git", "diff", "--exit-code", "-C", HOMEBREW_REPOSITORY, "manpages/brew.1"
+        test "git", "diff", "--exit-code", "-C", HOMEBREW_REPOSITORY, "manpages/brew-cask.1"
```
scpeters: Actually, I guess I could just get the `git diff` of the `manpages` folder and save a line

MikeMcQuaid: Maybe it's worth adding this inside `brew man` itself?

MikeMcQuaid:
Great idea, was on my list to check this out 👍

cmd/brew-test-bot.rb
scpeters: Interesting...call `git diff --exit-code` from `brew man` and return that exit code...that would be a nice place to put it.

scpeters: added in https://github.com/Homebrew/brew/pull/1177, I'll update this PR accordingly...

scpeters:
I updated to use the `--fail-if-changed` option instead of `--exit-code`.

Also I added a plain `brew man` test to improve coverage. This extra call adds another 2 seconds or so to test-bot.

cmd/brew-test-bot.rb
```diff
@@ -710,6 +710,11 @@ def homebrew
       return if @skip_homebrew
 
       if !@tap && (@formulae.empty? || @test_default_formula)
+        # this should always succeed
+        test "brew", "man"
```
MikeMcQuaid: Can remove the first call.

scpeters: done

PR #6, opened 2016-10-01
Run update-test --to-tag.

MikeMcQuaid:
Use this to ensure that tag updating isn't broken.


ilovezfs:
Looks like this broke CI.

PR #7, opened 2016-10-02
Use new `Formatter` module.

reitermarkus:


cmd/brew-test-bot.rb
```diff
@@ -179,19 +179,19 @@ def puts_command
         puts "travis_fold:start:#{@travis_fold_id}"
         puts "travis_time:start:#{@travis_timer_id}"
       end
-      puts "#{Tty.blue}==>#{Tty.white} #{@command.join(" ")}#{Tty.reset}"
+      puts Formatter.headline(@command.join(" "), color: :blue)
```
reitermarkus: Not sure about this one, as this is basically `ohai`.

MikeMcQuaid: I'm cool with using `ohai` if you think it makes more sense.

reitermarkus: You know, I still don't even know what `ohai` means, so let's leave it as-is for now. 😄

PR #16, opened 2016-10-22
Rubocop tweaks.

MikeMcQuaid:
Autofix some Homebrew Rubocop style complaints.


ilovezfs:
seems like the same style as brew repo should apply to custom commands ...

MikeMcQuaid:
Perhaps, yes.

MikeMcQuaid:
Yeh, got a Homebrew/brew PR incoming that will handle this.

PR #17, opened 2016-10-22
Use Rubocop to check homebrew/versions style.

MikeMcQuaid:
It passes `brew style` so whitelist it to be checked. This will allow more taps to be added in future.


ilovezfs:
maybe `STYLE_CHECKED_TAPS`? Well-styled sounds so sartorial.

MikeMcQuaid:
> Well-styled sounds so sartorial.

That was intentional 😆

PR #20, opened 2016-11-06
Test or_later formula changes.

MikeMcQuaid:
When the diff contains an `or_later` bottle change then don’t try to
build the bottle but just install it and test it that way.

CC @ilovezfs for thoughts.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -433,7 +433,15 @@ module Homebrew
       if @tap
         formula_path = @tap.formula_dir.to_s
         @added_formulae += diff_formulae(diff_start_sha1, diff_end_sha1, formula_path, "A")
-        @modified_formula += diff_formulae(diff_start_sha1, diff_end_sha1, formula_path, "M")\
+        @modified_formula += diff_formulae(diff_start_sha1, diff_end_sha1, formula_path, "M")
+        or_later_arg = "-G    sha256 ['\"][a-f0-9]*['\"] => :\w+_or_later$"
+        unless @modified_formula.empty?
+          unless git("diff", or_later_arg, "--unified=0").strip.empty?
+            # Test rather than build bottles if we're testing a `*_or_later`
+            # bottle change.
+            ARGV << "--no-bottle"
+          end
+        end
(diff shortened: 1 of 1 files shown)


ilovezfs:
It seems like it should be using json merges rather than manually editing bottle blocks, which we've emphasized as verboten.

MikeMcQuaid:
Can you elaborate more on "json merges"? In the short-term: do you object to this approach to verify if this is going to work for us? Happy to iterate on the workflow if it proves annoying but would be good to be able to use this for some more Sierra bottles temporarily.

ilovezfs:
I mean this jazz https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L968-L984 and the obverse in pull https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/pull.rb#L201-L218.

MikeMcQuaid:
Yeh, I see. In this case I think although it's different it's a feature because it doesn't require us to rebuild/reupload the bottle.

ilovezfs:
I don't think we'd need to rebuild/reupload to go through brew pull for this.

>In the short-term: do you object to this approach to verify if this is going to work for us?

As long as (A) we know the build fails on 10.12 and (B) the bottle goes through CI in some form before getting an or_later stamp, I think whatever works is fine in the short-run.

But I'd prefer to see these go through brew pull / git push, no upload/download of bottles involved, and the commit be attributed to test-bot not a maintainer.

MikeMcQuaid:
> As long as (A) we know the build fails on 10.12 and (B) the bottle goes through CI in some form before getting an or_later stamp, I think whatever works is fine in the short-run.

Yep, this will allow us to do this.



> But I'd prefer to see these go through brew pull / git push, no upload/download of bottles involved, and the commit be attributed to test-bot not a maintainer.

I'd rather avoid modifying e.g. `brew pull` for this weird case if not necessary.

ilovezfs:
>I'd rather avoid modifying e.g. brew pull for this weird case if not necessary.

I think it depends how many we're talking about. Given there are 65 left on https://github.com/Homebrew/homebrew-core/issues/6408 and maybe at most 1/4th of them actually need this right now, it's probably moot for the time being.

PR #21, opened 2016-11-06
Use uses --recursive.

MikeMcQuaid:
This was removed when we were stuck on Ruby 1.8 due to terrible performance but it should now be fine to re-enable.

CC @ilovezfs for thoughts.


ilovezfs:
I've always wanted this turned back on regardless of the performance implications, so :+1: :+1: 

We can always revert it if it becomes a problem.

MikeMcQuaid:
Glad we agree 👍

PR #22, opened 2016-11-06
Fix `or_later` detection.

MikeMcQuaid:
- escape `\` in regex to be passed to `git diff`
- specify the commit to check


ilovezfs:
looks like it works.

may want to tell `Warning: taylor has a bottle: do you need to update it with --bottle?` to go fly a kite

PR #24, opened 2016-11-20
brew-test-bot: Replace Utils::JSON with corelib JSON.

woodruffw:
ref: Homebrew/brew#1542


MikeMcQuaid:
Makes sense, thanks again @woodruffw!

PR #32, opened 2016-11-28
Fix pull --tap for Linuxbrew

sjackman:


sjackman:
Travis: `Error: GitHub API Error: API rate limit exceeded`
https://travis-ci.org/Homebrew/homebrew-test-bot/jobs/179314728#L930

cmd/brew-test-bot.rb
```diff
@@ -977,7 +980,10 @@ def test_ci_upload(tap)
 
     if (pr = ENV["UPSTREAM_PULL_REQUEST"])
       pull_pr = "https://github.com/#{tap.user}/homebrew-#{tap.repo}/pull/#{pr}"
-      safe_system "brew", "pull", "--clean", pull_pr
+      args = []
+      args << "--tap=#{@tap}" unless @tap.nil?
+      args << pull_pr
+      test "brew", "pull", "--clean", *args
```
MikeMcQuaid: Can you explain why both of these are needed, it's not clear? It seems like this should be autodetected by `brew pull` and probably fixed there. Additionally, this is a copy-paste so would be good to avoid duplication here.

sjackman: `Linuxbrew/science` is a fork of `Homebrew/science` used for building bottles without cluttering up the issue tracker of `Homebrew/science`. The URL of the repo is https://github.com/Linuxbrew/homebrew-science but the tap is `Homebrew/science`, as if you had done `brew tap homebrew/science https://github.com/Linuxbrew/homebrew-science`. It's tested using `brew test-bot --tap=homebrew/science`, and `test-bot` needs to pass the `--tap` parameter to `brew pull`, for example with `brew pull --tap=homebrew/science https://github.com/Linuxbrew/homebrew-science/pull/342`. I also use this arrangement to build bottles for `Homebrew/dupes`. That's the situation. Is what I've done the best way to fix this issue?

sjackman: The motivation is to build bottles using `test-bot` whose `INSTALL_RECEIPT.json` has `"tap":"homebrew/science"` rather than `"tap":"linuxbrew/science"` as if the following commands had been run:
```sh
brew tap homebrew/science https://github.com/Linuxbrew/homebrew-science
brew pull --tap=homebrew/science https://github.com/Linuxbrew/homebrew-science/pull/342
brew install --build-bottle bwa
brew bottle bwa
```

MikeMcQuaid: > brew pull --tap=homebrew/science https://github.com/Linuxbrew/homebrew-science/pull/342

That that works is at best an implementation detail and at worst a bug that we'll fix at a later point so I wouldn't rely on that behaviour.

> The motivation is to build bottles using test-bot whose INSTALL_RECEIPT.json has "tap":"homebrew/science" rather than "tap":"linuxbrew/science"

In that case you should adjust your bottle-building tooling to change the `INSTALL_RECEIPT.json` to produce the output you wish. It does seem, though, that having a dedicated Linuxbrew/homebrew-science tap is a much more sensible long-term decision. Again, that Homebrew/homebrew-science supports Linux is currently a lucky accident in which I'm currently looking the other way but the prevalence of `if OS.mac?` in that tap is not a good, maintainable long-term solution.

sjackman: A supported feature of `brew tap a/b http://github.com/c/homebrew-d` is that the git remote may be custom. When that's the case, to use the `brew pull` command, currently the tap needs to be specified, because it cannot be inferred from the pull request URL.

For example
```sh
brew tap a/b http://github.com/c/homebrew-d
brew pull --tap=a/b http://github.com/c/homebrew-d/pull/123
```

When the git remote of a tap is custom, how does one use the `brew pull` command with that tap?

sjackman: > In that case you should adjust your bottle-building tooling to change the `INSTALL_RECEIPT.json` to produce the output you wish.

Would you suggest adding a `--tap` option to the `brew install --build-bottle` command that is run by `brew test-bot`?

MikeMcQuaid: The custom argument option is to support non-GitHub or non-`homebrew-`-prefixed taps. In this case it is inferred but the result is not what you wish.



> Would you suggest adding a --tap option to the brew install --build-bottle command that is run by brew test-bot?

I would suggest you add it your e.g. CircleCI configuration.

sjackman: The Circle CI configuration runs `brew test-bot`, which creates the bottles. `brew install --build-bottle` creates the keg and `brew bottle` then creates the bottle. The `brew install` and `brew bottle` commands are inside `brew test-bot`, so the code to change the bottle tag of `INSTALL_RECEIPT.json` must live inside `brew test-bot`. I suggest using `inreplace` to change the `bottle` tag of `INSTALL_RECEIPT.json` between `brew install --bottle` and `brew bottle`. Does that work for you, or can you suggest an alternative solution?

MikeMcQuaid: My suggested solution would be to build bottles on Homebrew/homebrew-science if it's meant to support Linux (which I don't think is wise, as I've explained before, but I think that ship has probably sailed) or make Linuxbrew/linuxbrew-science a proper fork with the various Linux bottles in it.

sjackman: > My suggested solution would be to build bottles on Homebrew/homebrew-science if it's meant to support Linux

Can do. In fact I already have been doing for some time now. I originally built bottles on the Linuxbrew/science fork, but Linuxbrew/science can be decommissioned at this point. Close this issue?

PR #33, opened 2016-11-28
Fix --keep-old for Linuxbrew

sjackman:


sjackman:
Thanks for the merge, Mike!

PR #35, opened 2016-11-28
Use tap.remote for pull_pr for Linuxbrew

sjackman:


sjackman:
Thanks for the merge, Mike!

cmd/brew-test-bot.rb
```diff
@@ -976,7 +976,7 @@ def test_ci_upload(tap)
     safe_system "brew", "update"
 
     if (pr = ENV["UPSTREAM_PULL_REQUEST"])
-      pull_pr = "https://github.com/#{tap.user}/homebrew-#{tap.repo}/pull/#{pr}"
+      pull_pr = "#{tap.remote}/pull/#{pr}"
```
dunn: Would replacing `tap.remote` with `tap.remote.sub(/\.git$/, '')` fix this issue? https://bot.brew.sh/job/Homebrew%20Bottles/65369/console

sjackman: Yes, I think so.

PR #36, opened 2016-11-28
Fix test-bot --ci-upload for Linuxbrew

sjackman:


cmd/brew-test-bot.rb
```diff
@@ -1003,16 +1003,17 @@ def test_ci_upload(tap)
     bottles_hash.each do |formula_name, bottle_hash|
       version = bottle_hash["formula"]["pkg_version"]
       bintray_package = bottle_hash["bintray"]["package"]
+      bintray_org = OS.mac? ? "homebrew" : "linuxbrew"
```
MikeMcQuaid: Go for an `if`/`elsif` here just so we're not treating non-macOS as Linuxbrew.

sjackman: Fixed. Thanks for the suggestion, Mike.

sjackman:
Thanks for the merge, Mike.

PR #41, opened 2016-12-03
test-bot: use strict audits for updated formulae

jonchang:
Since the [pull request template](https://github.com/Homebrew/homebrew-core/blob/master/.github/PULL_REQUEST_TEMPLATE.md) asks contributors to run `brew audit --strict` it seems like the test-bot should do so too.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -515,7 +515,7 @@ module Homebrew
 
       new_formula = @added_formulae.include?(formula_name)
       audit_args = [formula_name]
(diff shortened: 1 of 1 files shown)


ilovezfs:
Basically the point of strict is to segregate the audit failures that should not cause test-bot to emit a CI failure, so this PR doesn't really make sense from that perspective. I think we throw the `--strict` in the template with the hope that some enterprising contributors will fix some of the more minor issues that the test-bot is *not* going to flag.

MikeMcQuaid:
Agreed with @ilovezfs but thanks for the PR anyway.

PR #50, opened 2017-02-05
test-bot: don't use git -C option

alyssais:
`-C` was only introduced in Git 1.8.5, which is older than the version of Git provided by some macOS versions and Linux distributions (including the CentOS 5 image used in homebrew-portable), so change the git command to use the equivalent options `--git-dir` and `--work-tree`, which have been around for much longer.

I don't know if this belongs here or in Linuxbrew's fork, since it's been a long time since macOS shipped with a git that old.


ilovezfs:
Linux fork IMO

alyssais:
Cool, I'll submit there.

PR #52, opened 2017-02-12
Don't require dependencies to be linked

MikeMcQuaid:
This should improve pull requests with lots of conflicting formulae in the same PR. Additionally, it’ll test that this linkage check isn’t required any longer.


MikeMcQuaid:
CI failing on Linux, will fix it later.

PR #53, opened 2017-02-26
Remove leftover Travis code

mistydemeo:
Since 4673c507, Travis no longer performs a `brew pull`; however, the URL to pull from is still being set even though it will never be used.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -355,8 +355,6 @@ module Homebrew
       @category = __method__
       @start_branch = current_branch
 
-      travis_pr = ENV["TRAVIS_PULL_REQUEST"] && ENV["TRAVIS_PULL_REQUEST"] != "false"
-
       # Use Jenkins GitHub Pull Request Builder plugin variables for
       # pull request jobs.
       if ENV["ghprbPullLink"]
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Just so I understand: does this manifest in any bug? Thanks!

mistydemeo:
It doesn't cause a bug, just confusion - on first read I was under the impression that this was still used, and it took some digging to find out why `brew test-bot` on Travis was tapping the master branch of a repo instead of the PR.

cmd/brew-test-bot.rb
```diff
@@ -369,10 +369,6 @@ def diff_formulae(start_revision, end_revision, path, filter)
         pr = $1
         @url = "#{git_url}/pull/#{pr}"
         @hash = nil
-      # Use Travis CI pull-request variables for pull request jobs.
-      elsif travis_pr
```
MikeMcQuaid: Can you remove the `travis_pr` variable that's set above and the `Travis` from `# Handle a URL being passed on the command-line or through Jenkins/Travis environment variables` below? Thanks ❤️

MikeMcQuaid:
@mistydemeo Gotcha, thanks.

cmd/brew-test-bot.rb
MikeMcQuaid: (Done)

mistydemeo:
Sorry for not getting to this sooner. 😰

PR #55, opened 2017-03-05
Only call `cask-tests` if the file exists.

reitermarkus:


MikeMcQuaid:
Now that https://github.com/Homebrew/brew/pull/2260 is merged can probably kill this altogether.

PR #58, opened 2017-03-08
Do not use ?override=1 when uploading to Bintray

sjackman:
The ?override=1 parameter is only needed to overwrite already published
artifacts. You can overwrite unpublished artifacts without it.


MikeMcQuaid:
👍 thanks @sjackman!

PR #60, opened 2017-04-11
test-bot: handle deleted formulae.

MikeMcQuaid:
When a formula is deleted the test bot run is currently a no-op. We may want to add more here in future
but to start with let's do a basic `brew uses` check to ensure no formulae still depend on it.

Also, while we're here: pluralise `@modified_formula` for consistency.

Requires https://github.com/Homebrew/brew/pull/2475 to work.

CC @ilovezfs for thoughts on any other omissions when we've deleted formulae.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -294,7 +294,8 @@ module Homebrew
       @url = nil
       @formulae = []
       @added_formulae = []
-      @modified_formula = []
+      @modified_formulae = []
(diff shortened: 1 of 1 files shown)


ilovezfs:
>CC @ilovezfs for thoughts on any other omissions when we've deleted formulae.

Possibly tap_migrations, formula_renames, broken aliases, and left over versioned formulae with no corresponding unversioned counterpart after the deletion.

PR #67, opened 2017-04-22
travis: cache gems, update Xcode.

MikeMcQuaid:
And change HOMEBREW_DEVELOPER assignment format to be the same as Homebrew/brew.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,19 @@
 language: ruby
+cache:
+  directories:
+    - $HOME/.gem/ruby
+    - Library/Homebrew/vendor/bundle
 
 matrix:
   include:
     - os: osx
-      osx_image: xcode8.1
+      osx_image: xcode8.3
       rvm: system
     - os: linux
       rvm: 2.0.0
 
 before_install:
-  - export HOMEBREW_DEVELOPER="1"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,15 +1,19 @@
...
     - os: osx
-      osx_image: xcode8.1
+      osx_image: xcode8.3
       rvm: system
     - os: linux
       rvm: 2.0.0
 
 before_install:
-  - export HOMEBREW_DEVELOPER="1"
+  - export HOMEBREW_DEVELOPER=1
```
reitermarkus: You could also use

```
env:
  - HOMEBREW_DEVELOPER=1
```

MikeMcQuaid: @reitermarkus Then Travis shows it in the UI and I get upset because I'm weird 😆. Thanks, though!

MikeMcQuaid: @reitermarkus Any other thoughts you have on improving our caching are super welcome, by the way.

reitermarkus: > Any other thoughts you have on improving our caching are super welcome, by the way.

Nothing I can think of.

> Then Travis shows it in the UI and I get upset because I'm weird 😆.

Ah, right, I know exactly what you mean! 😂

PR #76, opened 2017-05-05
test-bot: more conservatively cleanup etc/var.

MikeMcQuaid:
Instead of killing all of etc/var every time just delete the etc/var files that were installed from the bottle.


ilovezfs:
This seems to have brought back the original bug. In https://github.com/Homebrew/homebrew-core/pull/13274 the `mariadb` bottle has `.bottle/etc/my.cnf` and the `mysql` bottle does not.

MikeMcQuaid:
Fixed in https://github.com/Homebrew/homebrew-test-bot/commit/3c72d92cc5764aef25771ffd45b10cde0aba344d.

PR #81, opened 2017-05-09
Pass commit range to audit.

reitermarkus:
This passes the commit range of a PR to `audit`, used for the new check in https://github.com/Homebrew/brew/pull/2601.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -339,15 +339,37 @@ module Homebrew
       @repository.cd { Utils.popen_read("git", *args) }
     end
 
+    def shorten_revision(revision)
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Instead I'd rather we used the `FormulaVersions` class for this rather than modify `brew test-bot`.

reitermarkus:
We'll still need to know the commit range of a PR when using `FormulaVersions`, don't we?

MikeMcQuaid:
I'm not sure I understand why we would. We can detect `revision` and `version_scheme` changes without it so we should be able to do the same with a checksum.

reitermarkus:
Figured it out.

PR #88, opened 2017-05-19
Simplify `coverage_args`.

reitermarkus:
Maybe fixes https://github.com/Homebrew/brew/pull/2641.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -810,9 +810,7 @@ module Homebrew
     end
 
     def coverage_args
-      return [] unless ARGV.include?("--coverage")
-      return [] if @test_bot_tap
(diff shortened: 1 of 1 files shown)


cmd/brew-test-bot.rb
```diff
@@ -812,11 +812,7 @@ def deleted_formula(formula_name)
     def coverage_args
       return [] unless ARGV.include?("--coverage")
       return [] if @test_bot_tap
```
MikeMcQuaid: Let's remove this line. It's my fault rather than yours but this is why we didn't catch this before now and couldn't reproduce on this tap (which worried me).

MikeMcQuaid:
Great work here @reitermarkus 👏

PR #99, opened 2017-06-04
test-bot: run tests with --online.

MikeMcQuaid:
It's safe to assume given everything else that we have an internet connection for the bot.

Closes https://github.com/Homebrew/brew/pull/2691.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -836,9 +836,9 @@ module Homebrew
 
         test "brew", "style"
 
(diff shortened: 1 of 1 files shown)


JCount:
This failure seems to be caused by the fact that during the test, the branch git is on is `new branch` rather than master. This causes `expect { brew "pull", "https://github.com/Homebrew/homebrew-core/pull/1" }` to have non-standard behavior. The way to solve that would probably be to find a way to pass `--branch-ok` to `brew pull` as well.

PR #106, opened 2017-06-18
test-bot: manually run postinstall for devel.

ilovezfs:
If postinstall has not been run, some test blocks will fail. Unlike
stable, uninstalling and then pouring the devel bottle is not currently
an option because the mismatch between the bottle version and the
formula's stable version would trigger an exception. Just removing
--build-bottle for devel is also not an option since the .bottle
directory is needed for cleanup.


MikeMcQuaid:
Cheers for the PR!

PR #116, opened 2017-08-15
Use FormulaInstaller logic for satisfied reqs

### LINKED ISSUE #115: Dependency with xcode build requirement causes formula to be skipped
The `qt` formula has a [:build requirement on :xcode](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qt.rb#L29), but recently (possibly since #14), I am unable to build a bottle for a formula that depends on `qt` ( https://github.com/osrf/homebrew-simulation/pull/261 ) on a build machine that only has the command-line tools installed. The [build fails](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder-sierra/45/console) with the following console message:

~~~
==> SKIPPING: osrf/simulation/ignition-gui
A full installation of Xcode.app is required to compile this software.
Installing just the Command Line Tools is not sufficient.
Xcode can be installed from the App Store.
~~~

It should be fine to build without full Xcode since `qt` is being installed from a bottle. I can try to help with some guidance. I'm not sure if the problem is here or in `brew`.

scpeters:
Use FormulaInstaller::expand_requirements in test-bot's
satisfied_requirements? function to ensure that
build requirements are ignored for formulae that
are poured from bottles.
Fixes #115.

This solution has the advantage of de-duplicating some code between `brew` and `test-bot`, but it's possible that the logic in `FormulaInstaller::expand_requirements` is supposed to differ from the test-bot `satisfied_requirements?` function. If so, then this may not be the preferred solution. Thanks for your time.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -93,6 +93,7 @@
 #:    `GIT_URL`: if set to URL of a tap remote, same as `--tap`
 
 require "formula"
+require "formula_installer"
 require "utils"
 require "date"
 require "rexml/document"
@@ -552,14 +553,8 @@ module Homebrew
     end
 
     def satisfied_requirements?(formula, spec, dependency = nil)
-      requirements = formula.send(spec).recursive_requirements
-
-      unsatisfied_requirements = requirements.reject do |requirement|
-        satisfied = false
-        satisfied ||= requirement.satisfied?
-        satisfied ||= requirement.optional?
-        satisfied
-      end
+      fi = FormulaInstaller.new(formula)
+      unsatisfied_requirements, = fi.expand_requirements
 
       if unsatisfied_requirements.empty?
         true
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -553,7 +553,8 @@ module Homebrew
     end
 
     def satisfied_requirements?(formula, spec, dependency = nil)
-      fi = FormulaInstaller.new(formula)
+      f = Formulary.factory(formula.full_name, spec)
+      fi = FormulaInstaller.new(f)
       unsatisfied_requirements, = fi.expand_requirements
 
       if unsatisfied_requirements.empty?


ilovezfs:
how does this account for the difference between stable and devel?

ilovezfs:
FWIW: see my comment here: https://github.com/Homebrew/homebrew-core/pull/16205#issuecomment-318862727.

scpeters:
@ilovezfs oh, good point about stable and devel. Maybe it should be `FormulaInstaller.new(formula.send(spec))`? I'll see if that works too

scpeters:
Ok, I just tested it, and adding `.send(spec)` doesn't work. `FormulaInstaller` needs a `Formula` object, not `SoftwareSpec`. Does the `Formula` have an active spec embedded in it? I need to find a test case for this.

scpeters:
Maybe mongodb would be a good test case; it has a [devel build requirement on a specific Xcode](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mongodb.rb#L25-L29).

scpeters:
It's maybe not elegant, but I think I'm handling the spec properly now (1fab47975406e68c13001aa88fe4c29dcf99897c) by creating a new copy of the `Formula` object from `Formulary.factory` with the spec passed as an argument.

cmd/brew-test-bot.rb
```diff
@@ -552,18 +553,9 @@ def skip(formula_name)
...
-        satisfied ||= requirement.optional?
-        if !satisfied && requirement.default_formula?
-          default = Formula[requirement.default_formula]
-          satisfied = satisfied_requirements?(default, :stable, formula.full_name)
-        end
-        satisfied
-      end
+      f = Formulary.factory(formula.full_name, spec)
+      fi = FormulaInstaller.new(f)
+      unsatisfied_requirements, _ = fi.expand_requirements
```
MikeMcQuaid: You can skip the `_` here, by the way.

Feels a little weird using `FormulaInstaller` here. What do you (and @ilovezfs) think about maybe adding/moving the `expand_requirements` logic into `Formula`? I haven't looked at how gnarly that would be. If it's hard: this seems fine.

@ilovezfs any other thoughts here?

scpeters: I agree that it feels weird. The reason I used `FormulaInstaller` was for the `install_bottle_for?` logic, which helps determine when non-runtime requirements aren't necessary ([this logic branch specifically](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula_installer.rb#L446-L447)).

ilovezfs:
@scpeters wouldn't it be sufficient to skip optional and build dependencies in https://github.com/Homebrew/brew/blob/master/Library/Homebrew/software_spec.rb#L173-L190 and not change anything in test-bot?

scpeters:
@ilovezfs is it valid to skip build dependencies in `software_spec::recursive_dependencies`? It would depend on how that function is used elsewhere, I suppose. What if you are building a bottle for that formula so you need its build dependencies?

scpeters:
> wouldn't it be sufficient to skip optional and build dependencies in https://github.com/Homebrew/brew/blob/master/Library/Homebrew/software_spec.rb#L173-L190 and not change anything in test-bot?

Coming back to this, I'm not sure that exact code block from `software_spec` is relevant since it addresses dependencies, but this PR is focused on requirements. Do you propose modifying the `recursive_requirements` function in `software_spec`? If you have an alternative patch, I would be happy to test it; in the meantime, I'm not sure how to test your idea.

ilovezfs:
No, recursive_requirements relies on expand, which goes through the dependencies.

ilovezfs:
See https://github.com/Homebrew/brew/pull/2822 which was added specifically to support the code here.

scpeters:
Is this what you had in mind?

~~~
diff --git a/Library/Homebrew/software_spec.rb b/Library/Homebrew/software_spec.rb
index dd6026fcf4..9ad9f0ab1a 100644
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -194,7 +194,10 @@ class SoftwareSpec
   end
 
   def recursive_requirements
-    Requirement.expand(self)
+    Requirement.expand(self) do |_, requirement|
+      Requirement.prune if requirement.build?
+      Requirement.prune if !requirement.required? && build.without?(requirement)
+    end
   end
 
   def patch(strip = :p1, src = nil, &block)
~~~

This patch doesn't appeal to me because while it would probably fix #115, it seems like it would render all `:build` requirements toothless (for example when building `qt` next, would the `:Xcode` requirement be honored?).

ilovezfs:
I don't know. You would need to test.

As I said, the recursive dependencies section not the recursive requirements section.

In any case, I'm :-1: on this PR because this seems like an abuse of FormulaInstaller and not something that belongs in test-bot.

MikeMcQuaid:
> This patch doesn't appeal to me because while it would probably fix #115, it seems like it would render all :build requirements toothless (for example when building qt next, would the :Xcode requirement be honored?).

@scpeters Can this be modified to take that sort of approach instead? Can you elaborate on what you mean and why you feel this approach here is superior (e.g. does this approach take into account what's needed for bottled/unbottled deps?).

scpeters:
In #115, `qt`'s build requirement on xcode is being propagated to its dependents even if qt itself is being installed from a bottle. I haven't tested it, but my worry was that the software_spec patch would make all `:build` requirements be ignored, which seems like throwing the baby out with the bathwater.

I'll close this for now while I work on testing the software_spec patch. Thanks for your time and reviews.

MikeMcQuaid:
Thanks for understanding @scpeters!

scpeters:
I appreciate the work that you do and try to help in small ways when I can.

MikeMcQuaid:
@scpeters If you undelete the branch we can reopen this and figure out to improve the Homebrew/brew APIs this uses.

scpeters:
I force-pushed to this branch, so it doesn't want to let me reopen; I'll create a new PR

ilovezfs:
if you force push to a8b24646fed9a32ef3a08405f71195f1cbae21ab it should let you reopen

scpeters:
ok that was tricky; I'll rebase now that it's open again

scpeters:
what an adventure

ilovezfs:
I think this is still going to be buggy due to https://github.com/Homebrew/brew/issues/3615 but hopefully an improvement on the status quo.

MikeMcQuaid:
Thanks again @scpeters!

ilovezfs:
It looks like there's a minor bug here:
https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/17524/version=el_capitan/console

```
+ brew test-bot --ci-pr
Homebrew/homebrew-test-bot 7bab170 (Merge pull request #116 from scpeters/expand_req)
ARGV: --ci-pr --cleanup --test-default-formula --local --junit
==> git -C homebrew/homebrew-core checkout origin/master
==> brew pull --clean https://github.com/Homebrew/homebrew-core/pull/23483
Homebrew/brew 1.5.2-35-gc77cafc (Merge pull request #3729 from ilovezfs/bump-formula-pr-more-enhancements)
homebrew/core origin/master 655ff26 (rabbitmq 3.7.3 (#23466))
homebrew/core HEAD 117b81a (stress-ng 0.09.14)
==> brew doctor
==> brew --env
==> brew config
==> brew readall --aliases homebrew/core
==> brew uses --recursive stress-ng
==> SKIPPING: stress-ng
==> git -C homebrew/homebrew-core checkout master -f
==> brew cleanup --prune=7
Error: undefined method `message' for #<Array:0x0000010291a5c0>
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/cmd/brew-test-bot.rb:567:in `each'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/cmd/brew-test-bot.rb:567:in `map'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/cmd/brew-test-bot.rb:567:in `satisfied_requirements?'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/cmd/brew-test-bot.rb:682:in `formula'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/cmd/brew-test-bot.rb:1113:in `block in run'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/cmd/brew-test-bot.rb:1112:in `each'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/cmd/brew-test-bot.rb:1112:in `run'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/cmd/brew-test-bot.rb:1438:in `test_bot'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/cmd/brew-test-bot.rb:1526:in `<top (required)>'
/usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.3/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.3/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/usr/local/Homebrew/Library/Homebrew/utils.rb:20:in `require?'
/usr/local/Homebrew/Library/Homebrew/brew.rb:106:in `<main>'
Build step 'Execute shell' marked build as failure
Recording test results
ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?
Archiving artifacts
[WS-CLEANUP] Deleting project workspace...[WS-CLEANUP] done
Finished: FAILURE
```

ilovezfs:
Yeah same thing again in another PR https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/17532/version=el_capitan/console

scpeters:
I think I mixed up the data structures; the previous `unsatisfied_requirements` was a Hash, but now it is a set. I should have tested it on something that had unsatisfied requirements.

scpeters:
~~~
diff --git a/cmd/brew-test-bot.rb b/cmd/brew-test-bot.rb
index e6c1f09..2c0d7c0 100755
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -564,7 +564,7 @@ module Homebrew
         name += " (#{spec})" unless spec == :stable
         name += " (#{dependency} dependency)" if dependency
         skip name
-        puts unsatisfied_requirements.map(&:message)
+        puts unsatisfied_requirements.values.map(&:message)
         false
       end
     end
~~~

ilovezfs:
:tada:

scpeters:
Does that patch work for you?

scpeters:
I submitted that patch in #138

ilovezfs:
looks like not
```
==> SKIPPING: carthage
Error: undefined method `message' for #<Array:0x00000102011ac8>
```
old error ^^


```
==> SKIPPING: carthage
Error: undefined method `message' for [#<XcodeRequirement: "xcode" [:build] version="9.0">]:Array
```
new error ^^

scpeters:
:cry:

ilovezfs:
I think we want `puts unsatisfied_requirements.values.flatten.map(&:message)`

scpeters:
11aa8e1fd3ad4b7a92f46c110045d8afe0b55adc in #138

scpeters:
oh, that works too, I did `.first`, but I'll switch to `.flatten`

scpeters:
4ac0211bb87323819b4fcb7c74d88c92a25e9e35 uses flatten

PR #91, opened 2017-05-24
Tap if necessary during unlink_conflicts

scpeters:
Related to #86, rescue from TapFormulaUnavailableError
during call to formula.recursive_dependencies
in unlink_conflicts (called by install_bottled_dependent)
and add the missing tap.

I found #86 when trying to `test-bot` a formula (ie. `osrf/simulation/gazebo7`) that depends on an untapped formula (ie. `cartr/qt4/qt@4`). For this PR, I found a problem when rebuilding a bottle for `ignition-transport`, which is a dependency of `gazebo7`. When trying to `install_bottled_dependent`, there was a call to `formula.recursive_dependencies` in `unlink_conflics` ([stacktrace here](https://gist.github.com/scpeters/7b3f46e672cd126507e06aab19e99030)).

PR #117, opened 2017-08-28
Check for untapped deps when sorting formulae

scpeters:
I maintain at least one formula with cross-tap dependencies ([osrf/simulation/gazebo7](https://github.com/osrf/homebrew-simulation/blob/master/gazebo7.rb#L29) depends on `cartr/qt4/qt@4`). In #91, I fixed a bug in `unlink_conflicts` related to this use case that rescues from `TapFormulaUnavailableError`, adds the missing tap, and tries again to call `recursive_dependencies` on a formula.

I just noticed there is an earlier place where a missing tap causes a problem: in the `run` function, a [formulae.each loop](https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L1105-L1111) is right after `homebrew`. 
The `formulae` function [calls brew deps](https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L1080) and then sorts the formulae so the lowest level dependencies are processed first, but `brew deps` fails if untapped formulae aren't available:

~~~
$ brew deps gazebo7 && echo success
Error: No available formula with the name "cartr/qt4/qt@4" 
Please tap it and then try again: brew tap cartr/qt4
$
~~~

I hadn't noticed this before when just building a gazebo7 bottle, but I now notice that the order is wrong when trying to build bottles for https://github.com/osrf/homebrew-simulation/pull/269, which has bumped revisions for `gazebo7` and two of its dependencies that are later in the alphabet, so it [builds out of order](https://build.osrfoundation.org/view/os_homebrew/job/generic-release-homebrew_bottle_builder-elcapitan/64/console).

This PR checks `$?` to see if the `brew deps` command failed, and if so calls `Formulary.factory(formula).recursive_dependencies` with a `rescue` block to add the missing tap. I put WIP in the title because it feels messy to duplicate code from #91. I am assuming that `brew deps` behavior should not change to automatically add missing taps.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -1085,7 +1085,18 @@ module Homebrew
       changed_formulae_dependents = {}
 
       @formulae.each do |formula|
-        formula_dependencies = Utils.popen_read("brew", "deps", "--full-name", "--include-build", formula).split("\n")
+        begin
+          formula_dependencies = Utils.popen_read("brew", "deps", "--full-name", "--include-build", formula).split("\n")
+          # deps can fail if deps are not tapped
+          unless $?.success?
+            Formulary.factory(formula).recursive_dependencies
+          end
+        rescue TapFormulaUnavailableError => e
+          raise if e.tap.installed?
+          e.tap.clear_cache
+          safe_system "brew", "tap", e.tap.name
+          retry
+        end
         unchanged_dependencies = formula_dependencies - @formulae
         changed_dependences = formula_dependencies - unchanged_dependencies
         changed_dependences.each do |changed_formula|


MikeMcQuaid:
Seems reasonable, thanks @scpeters. @ilovezfs any thoughts?

scpeters:
I removed `WIP` from the title because I think it's ready for review.

scpeters:
I just reconfirmed the bug and fix with https://github.com/osrf/homebrew-simulation/pull/372

MikeMcQuaid:
Thanks again @scpeters!

PR #124, opened 2017-10-27
test-bot: make some style fixes.

MikeMcQuaid:
Flagged by RuboCop.


ilovezfs:
This causes all bottle jobs to fail.

PR #132, opened 2018-01-12
Add support for xz bottles

kabel:
* Installs XZ during setup
* Finds the json file from xz bottles

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -620,8 +620,10 @@ module Homebrew
       return if ARGV.include?("--fast")
       return if ARGV.include?("--no-bottle")
       return if formula.bottle_disabled?
-      xz_formula = Formulary.factory("xz")
-      test "brew", "install", "xz" unless xz_formula.installed?
+      if OS.mac?
+        xz_formula = Formulary.factory("xz")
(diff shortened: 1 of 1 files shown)


kabel:
Related to https://github.com/Homebrew/brew/pull/3641

cmd/brew-test-bot.rb
```diff
@@ -583,6 +583,7 @@ def setup
       test "brew", "doctor"
       test "brew", "--env"
       test "brew", "config"
+      test "brew", "install", "xz"
```
MikeMcQuaid: This should be conditional on if it's already installed, if `!ARGV.include?("--no-bottle")` and just before the first bottle is created https://github.com/kabel/homebrew-test-bot/blob/a8e82986e8ecbe70a02ddf540aaefbf7720afff9/cmd/brew-test-bot.rb#L628

kabel:
Any new feedback or help on how to fix linux not being able to install xz?

MikeMcQuaid:
Add an `if OS.mac?` or make it conditional on `Formula["xz"]` returning a value rather than throwing an exception.

MikeMcQuaid:
Passing on this due to reasons explained in https://github.com/Homebrew/brew/pull/3641#issuecomment-363020837. Please take any conversation there.

PR #141, opened 2018-02-21
Minor issues

adius:
Hi,

this pull request was created with help of [Feram]!
It's a tool which automatically finds and fixes bugs in your repos.

To view the other patches we created for your code
simply activate [Feram] on [github.com/apps/feram].
Or if you are new to GitHub apps,
check out the full tutorial at [feram.io/help/getting-started].

If you have any feedback or wishes, don't hesitate to contact me
or open an issue at [github.com/feramhq/feram]!

Hope you like it! 😁

[feram.io/help/getting-started]: https://feram.io/help/getting-started
[Feram]: https://www.feram.io
[github.com/apps/feram]: https://github.com/apps/feram
[github.com/feramhq/feram]: https://github.com/feramhq/feram

### DIFF
--- a/Formula/testbottest.rb
+++ b/Formula/testbottest.rb
@@ -1,21 +1,21 @@
 class Testbottest < Formula
-  desc "Minimal C program and Makefile used for testing Homebrew"
-  homepage "https://github.com/Homebrew/brew"
-  url "file://#{File.expand_path("..", __FILE__)}/tarballs/testbottest-0.1.tbz"
-  sha256 "246c4839624d0b97338ce976100d56bd9331d9416e178eb0f74ef050c1dbdaad"
-  head "https://github.com/Homebrew/homebrew-test-bot.git"
+  desc 'Minimal C program and Makefile used for testing Homebrew'
+  homepage 'https://github.com/Homebrew/brew'
+  url "file://#{File.expand_path('..', __FILE__)}/tarballs/testbottest-0.1.tbz"
+  sha256 '246c4839624d0b97338ce976100d56bd9331d9416e178eb0f74ef050c1dbdaad'
+  head 'https://github.com/Homebrew/homebrew-test-bot.git'
 
-  depends_on :java => ["1.0+", :optional]
+  depends_on java: ['1.0+', :optional]
 
   fails_with :gcc do
     build 1
-    cause "testing"
+    cause 'testing'
   end
 
   def install
     odie "whoops, shouldn't be using java!" if build.with?(:java)
 
-    system "make", "install", "PREFIX=#{prefix}"
+    system 'make', 'install', "PREFIX=#{prefix}"
   end
 
   def post_install


MikeMcQuaid:
This change is incorrect and makes CI fail.

adius:
Sorry. It seems like Homebrew is the only larger Ruby project which does not use the Rails styleguide 🙄

MikeMcQuaid:
@adius We use `rubocop` to enforce our code style consistently through the Homebrew ecosystem and this is enforced through tests. I'm 100% certain you will find other large Ruby projects that don't follow the Rails style guide and if you're going to submit automated PRs to them you should probably check if you're going to fail CI before you do so. Also, I'm going to try and assume the 🙄 wasn't directed at this project or me because if it was it would be extremely rude.

adius:
> We use rubocop to enforce our code style consistently through the Homebrew ecosystem

But there is no config Feram could have automatically picked up, is there?

> should probably check if you're going to fail CI before you do so

Good point. I'll investigate what's possible in that direction, cause I think I have to open a PR to trigger the CI. So there's no way to check it before (I guess we could at least automatically close the PR)

> 🙄 wasn't directed at this project or me because if it was it would be extremely rude.

No, wasn't. Rather at the state of software development ... it's a mess ... although not anyone's fault really ^^.

scpeters:
I believe the relevant rubocop file is in brew:

* https://github.com/Homebrew/brew/blob/master/Library/.rubocop.yml#L193-L199

MikeMcQuaid:
> I'll investigate what's possible in that direction, cause I think I have to open a PR to trigger the CI. So there's no way to check it before (I guess we could at least automatically close the PR)

You could read from a `.travis.yml` file.

> I believe the relevant rubocop file is in brew:
> 
> https://github.com/Homebrew/brew/blob/master/Library/.rubocop.yml#L193-L199

Indeed. This saves duplication in every tap.

adius:
> You could read from a .travis.yml file.

But than we'd have to run the tests ourselvs ... so basically rebuild the travis infrastructure ... which is kind of hard 😅

> I believe the relevant rubocop file is in brew:

Ok, I see, but this means it's basically impossible to pick it up automatically.

> this PR feels like it has a financial motivation to promote your business. 

I think they're getting it all wrong. My motivation is to help developers write better code and save them from stupid obvious bugs which cost them time and get on their nerves. The company is just the vehicle to do this. (and at the moment it's unfortunately more about staying alive than anything else 😞)

> My comments and advice above shouldn't be just seen at me being irritated at a bad PR but trying to give you advice so other ecosystems don't get negative associations with Feram.

And I'm really glad about any feedback! It's really hard to automatically detect all the custom preferences and honor them in the code changes, but we're getting better at it each week.

adius:
But most of all it's my fault. I checked those changes and they looked legit to me, but yeah obviously they aren't...

PR #151, opened 2018-03-31
Add option --root-url

sjackman:
If `--root-url` is passed, use the specified URL as the root of the bottle's URL instead of Homebrew's default.


MikeMcQuaid:
Seems reasonable, thanks @sjackman!

sjackman:
Thanks for merging, Mike! 👍

PR #152, opened 2018-04-11
Add option --or-later

sjackman:
If `--or-later` is passed, append `_or_later` to the bottle tag.


MikeMcQuaid:
Seems reasonable, thanks again @sjackman!

sjackman:
Thanks again for merging, Mike!

PR #153, opened 2018-04-24
test-bot: only run setup once

ilovezfs:
Avoid strange broken opt symlink errors on some `brew doctor` runs other
than the first.


MikeMcQuaid:
Thanks @ilovezfs.

PR #164, opened 2018-07-04
testbottest: add bottle block

commitay:
This adds a bottle block to satisfy the `audit --online` step of the travis `brew` tests.

https://github.com/Homebrew/brew/pull/4420

### DIFF
--- a/Formula/testbottest.rb
+++ b/Formula/testbottest.rb
@@ -5,6 +5,9 @@ class Testbottest < Formula
   sha256 "246c4839624d0b97338ce976100d56bd9331d9416e178eb0f74ef050c1dbdaad"
   head "https://github.com/Homebrew/homebrew-test-bot.git"
 
+  bottle do
+  end
+
   depends_on :java => ["1.0+", :optional]
 
   fails_with :gcc do


MikeMcQuaid:
Maybe make this bottle disabled or unneeded instead? Want to ensure it’s always built from source for testing.

ilovezfs:
Maybe could add a `--force` to `brew bottle`?

MikeMcQuaid:
What if this is bottle do with nothing else in it?

commitay:
Empty `bottle do` block seems to work but it errors with `--strict` for [bottle_block_cop.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/rubocops/bottle_block_cop.rb). Not sure if that's an issue or not.

```
An error occurred while FormulaAuditStrict/BottleBlock cop was inspecting ... testbottest
undefined method `metadata' for nil:NilClass
```

commitay:
Pushed an empty block to see if CI likes it or not.

MikeMcQuaid:
Thanks again @commitay!

PR #166, opened 2018-07-05
Revert "testbottest: add bottle block"

commitay:
Apologies, this doesn't actually work. Reverting in favour of exempting the `testbottest` formula from the bottle audit in https://github.com/Homebrew/brew/pull/4420.


MikeMcQuaid:
👍 thanks for following up.

PR #175, opened 2018-08-23
brew-test-bot: check build/optional for deleted formulae

commitay:
There was an expectation https://github.com/Homebrew/homebrew-core/pull/31382 should be red-flagged by CI because that formula is currently a mandatory dependency, albeit at buildtime, of `elm` and consequently removing it would break `elm`, but instead CI failed to provide any warning about consequences of removing it.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -982,6 +982,7 @@ module Homebrew
       @category = "#{__method__}.#{formula_name}"
       test "brew", "uses", "--include-build",
                            "--include-optional",
+                           "--include-test",
                            formula_name
     end
 
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -980,7 +980,9 @@ module Homebrew
 
     def deleted_formula(formula_name)
       @category = "#{__method__}.#{formula_name}"
-      test "brew", "uses", formula_name
+      test "brew", "uses", "--include-build",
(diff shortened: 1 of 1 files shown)


DomT4:
I don't care _as much_ about catching `:optional` stuff in advance of removal because generally those should either be on by default or safe enough to remove, when the problem is caught, if the optional dependency goes away for whatever reason, but `:build` is definitely equivalent to being a mandatory dependency at least for CI purposes.

A definite 👍 to `:build` and a 👍 to `:optional` if nobody raises an objection.

MikeMcQuaid:
> No description provided.

Can you provide a bit more context here? Thanks.

DomT4:
There was an expectation https://github.com/Homebrew/homebrew-core/pull/31382 should be red-flagged by CI because that formula is currently a mandatory dependency, albeit at buildtime, of `elm` and consequently removing it would break `elm`, but instead CI failed to provide any warning about consequences of removing it.

I think we should cover at least both `:build` and `:test` dependencies on the deleted formula check given both are essentially mandatory dependencies for CI/contributor/etc purposes.

MikeMcQuaid:
Thanks for the context, @DomT4.

> Check if a deleted formulae is used as a build time or optional dependency by other formulae.

@commitay FYI that describes _what_ this change is but not _why_.

MikeMcQuaid:
Thanks again @commitay!

PR #177, opened 2018-08-25
Update cache link logic.

reitermarkus:
Depends on https://github.com/Homebrew/brew/pull/4662.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -775,7 +775,17 @@ module Homebrew
       test "brew", "bottle", *bottle_merge_args
       test "brew", "uninstall", "--force", formula.name
 
-      FileUtils.ln bottle_filename, HOMEBREW_CACHE/bottle_filename, force: true
+      bottle_json = JSON.parse(File.read(bottle_json_filename))
+      root_url = bottle_json.dig(formula.full_name, "bottle", "root_url")
+      filename = bottle_json.dig(formula.full_name, "bottle", "tags").values.first["filename"]
+
+      download_strategy = CurlDownloadStrategy.new("#{root_url}/#{filename}", formula.name, formula.version)
+
+      FileUtils.ln bottle_filename, download_strategy.cached_location, force: true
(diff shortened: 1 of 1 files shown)


cmd/brew-test-bot.rb
```diff
@@ -775,7 +775,15 @@ def bottle_reinstall_formula(formula, new_formula)
...
 
-      FileUtils.ln bottle_filename, HOMEBREW_CACHE/bottle_filename, force: true
+      bottle_json = JSON.parse(File.read(bottle_json_filename))
+
+      root_url = bottle_json.dig(formula.name, "bottle", "root_url")
+      filename = bottle_json.dig(formula.name, "bottle", "tags").values.first["filename"]
+      url_sha256 = Digest::SHA256.hexdigest("#{root_url}/#{filename}")
+
+      FileUtils.ln bottle_filename, HOMEBREW_CACHE/"downloads/#{url_sha256}--#{filename}", force: true
+      FileUtils.ln_s "downloads/#{url_sha256}--#{filename}", HOMEBREW_CACHE/bottle_filename, force: true
```
MikeMcQuaid: Could this maybe get refactored into some method (doesn't have to be public) in Homebrew/brew that test-bot can call instead? Don't really like the logic being duplicated here.

reitermarkus: Any suggestion on where to put it? I thought about it but I can't think of a good place where this logic should go. Even in the download strategies, these steps don't happen in a single cohesive action, and `FileUtils.ln` is specific to `test-bot`.

reitermarkus:
@MikeMcQuaid, I changed it to used the `CurlDownloadStrategy` strategy to get the proper name.

Instead of the `FileUtils.ln_s` step, we could now also use `download_strategy.fetch`, which will show “Already downloaded” and create the symlink.

MikeMcQuaid:
> Instead of the FileUtils.ln_s step, we could now also use download_strategy.fetch, which will show “Already downloaded” and create the symlink.

This works 👍

PR #178, opened 2018-09-06
test-bot: install newer Git when needed.

MikeMcQuaid:
We're relying on newer Git syntax now.


MikeMcQuaid:
@fxcoudert Trying to update in https://github.com/Homebrew/homebrew-test-bot/pull/180.

PR #189, opened 2018-09-18
Remove `run_as_not_developer` calls.

reitermarkus:
Needed for https://github.com/Homebrew/brew/pull/4909.


MikeMcQuaid:
@reitermarkus There's a bunch of other changes here, can you explain them?

reitermarkus:
```ruby
$stdout.sync = true
$stderr.sync = true
```

ensures the output is unbuffered to avoid weird output on Travis. Everything else is just refactoring.

PR #198, opened 2018-10-03
test-bot: check shell files with shellcheck.

MikeMcQuaid:
This will do a similar job to RuboCop for us.


MikeMcQuaid:
Will move this into `brew style` instead.

PR #203, opened 2018-10-03
testbottest: Rename :gcc to :gcc_4_2

sjackman:


MikeMcQuaid:
Thanks @sjackman!

sjackman:
Thanks for merging, Mike!

PR #205, opened 2018-10-11
azure/travis: update to xcode 10

commitay:


MikeMcQuaid:
@commitay Why the close?

commitay:
When I opened this I'd forgotten about Azure. They have Xcode 10 support but
`vmImage: xcode10-macos10.13` doesn't seem to work.

MikeMcQuaid:
@commitay Do you have a failed log or similar you can share? Would be good to report that to MS.

commitay:
https://dev.azure.com/Homebrew/Homebrew/_build/results?buildId=484&view=results

screen shot 2018-10-11 at 23 11 37

MikeMcQuaid:
@commitay Thanks! Looks like there may be an Xcode 10 in the image we're currently using if we `xcode-select --switch` to it?

commitay:
Seems to be working.

MikeMcQuaid:
@commitay 👍 merge when ready (and open a similar PR for Homebrew/brew, thanks 🙇)

PR #219, opened 2018-12-30
brew-test-bot.rb: tweak missing bottles message when not using jenkins

iMichka:
When not using jenkins (for example when using AWS lambda for linuxbrew,
or even when manually uploading a bottle), the current error message
was "Missing Jenkins variables" when there were no bottles in the dir.
This may happen when no bottle has been built by the CI, or one
is trying to upload bottles from the wrong folder.

The new error messages have been improved and the path is also
displayed, to ease debugging.

There is one assumption here: when JENKINS_HOME is set, one has
to retrieve the bottles first (previously this code path
was triggered by the fact that there were no bottles in the
starting repository)

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -1308,6 +1308,23 @@ module Homebrew
     end
   end
 
+  def copy_bottles_from_jenkins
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @iMichka!

PR #223, opened 2019-01-07
WIP brew-ci-upload: migrate upload code to separate command

iMichka:
Currently, the brew-test-bot.rb file is 1748 lines long. This is way too long in my opinion.

I always found weird that the CI upload code was part of the brew-test-bot file, so moving that part out of the file would allow to trim down a little bit the code.

This is a WIP: I would like to introduce a new command, `brew ci-upload`, which allows to upload bottles to bintray. The current code is just a copy-paste from the code initially in `brew-test-bot.rb`.

The migration plan would be the following:
- Merge this once it is ready
- Update the callers to use this new command
- Delete the old code from `brew-test-bot.rb`

Any thoughts?

I also need help for one thing: the `resolve_test_tap` function is used by both files: where should I put it? In a shared file at the same level as the cmd folder, or in some library subfolder? What is the best practice in ruby?

### DIFF
--- a/cmd/brew-ci-upload.rb
+++ b/cmd/brew-ci-upload.rb
@@ -1,0 +1,314 @@
+#:  * `ci-upload` [options]  <url|formula>:
+#:    Tests the full lifecycle of a formula or Homebrew/brew change.
+#:
+#:    If `--dry-run` is passed, print what would be done rather than doing
+#:    it.
+#:
+#:    If `--bintray-org=<bintray-org>` is passed, upload to the given Bintray
+#:    organisation.
+#:
+#:    If `--tap=<tap>` is passed, use the `git` repository of the given
+#:    tap.
+#:
+#:    If `--git-name=<git-name>` is passed, set the Git
+#:    author/committer names to the given name.
+#:
+#:    If `--git-email=<git-email>` is passed, set the Git
+#:    author/committer email to the given email.
+
+
+module Homebrew
+  module_function
+
+  def ci_upload
+      $stdout.sync = true
+      $stderr.sync = true
+
+      tap = resolve_test_tap
+      # Tap repository if required, this is done before everything else
+      # because Formula parsing and/or git commit hash lookup depends on it.
+      # At the same time, make sure Tap is not a shallow clone.
+      # bottle rebuild and bottle upload rely on full clone.
+      if tap
+        if !tap.path.exist?
+          safe_system "brew", "tap", tap.name, "--full"
+        elsif (tap.path/".git/shallow").exist?
+          raise unless quiet_system "git", "-C", tap.path, "fetch", "--unshallow"
+        end
+      end
+
+      # Don't trust formulae we're uploading
+      ENV["HOMEBREW_DISABLE_LOAD_FORMULA"] = "1"
+
+      bintray_user = ENV["HOMEBREW_BINTRAY_USER"]
+      bintray_key = ENV["HOMEBREW_BINTRAY_KEY"]
+      if !bintray_user || !bintray_key
+        unless ARGV.include?("--dry-run")
+          raise "Missing HOMEBREW_BINTRAY_USER or HOMEBREW_BINTRAY_KEY variables!"
+        end
+      end
+
+      # Ensure that uploading Homebrew bottles on Linux doesn't use Linuxbrew.
+      bintray_org = ARGV.value("bintray-org") || "homebrew"
+      if bintray_org == "homebrew" && !OS.mac?
+        ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1"
+      end
+
+      # Don't pass keys/cookies to subprocesses
+      ENV.clear_sensitive_environment!
+
+      ARGV << "--verbose"
+
+      copy_bottles_from_jenkins if !ENV["JENKINS_HOME"].nil?
+
+      raise "No bottles found in #{Dir.pwd}!" if Dir["*.bottle*.*"].empty? && !ARGV.include?("--dry-run")
+
+      json_files = Dir.glob("*.bottle.json")
+      bottles_hash = json_files.reduce({}) do |hash, json_file|
+        hash.deep_merge(JSON.parse(IO.read(json_file)))
+      end
+
+      if ARGV.include?("--dry-run")
+        bottles_hash = {
+          "testbottest" => {
+            "formula" => {
+              "pkg_version" => "1.0.0",
+            },
+            "bottle"  => {
+              "rebuild" => 0,
+              "tags"    => {
+                Utils::Bottles.tag => {
+                  "filename" =>
+                                "testbottest-1.0.0.#{Utils::Bottles.tag}.bottle.tar.gz",
+                  "sha256"   =>
+                                "20cdde424f5fe6d4fdb6a24cff41d2f7aefcd1ef2f98d46f6c074c36a1eef81e",
+                },
+              },
+            },
+            "bintray" => {
+              "package"    => "testbottest",
+              "repository" => "bottles",
+            },
+          },
+        }
+      end
+
+      first_formula_name = bottles_hash.keys.first
+      tap_name = first_formula_name.rpartition("/").first.chuzzle
+      tap_name ||= CoreTap.instance.name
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'd rather the old code was deleted at the same time and the `brew test-bot --ci-upload` flag remains but calls into the next script until callers are updated.



> I also need help for one thing: the `resolve_test_tap` function is used by both files: where should I put it?

I think this is an argument for keeping it in the same file, honestly. I'm not sure it's worth splitting up what is essentially just a CI script. Instead of moving into multiple files what about splitting into more methods or modules?

iMichka:
> I'd rather the old code was deleted at the same time and the brew test-bot --ci-upload flag remains but calls into the next script until callers are updated.

I'm ok with that strategy too.

> I think this is an argument for keeping it in the same file, honestly.

I feel that the `brew-test-bot.rb` file has grown too much. It is daunting to read, I often get lost in it when looking for bugs. You need to scroll a lot around when debugging stuff, which is for me a sign that something is wrong. Searching for text in it makes you jump around, but as the script is long, you have no clue in which part of the code you are. Of course I should spend more time reading it, but splitting it would make me read just the part I am interested in.

The script does 3 different things from a functional point of view (if I did not miss other things): test brew, build bottles, upload bottles.

This change is the first step: I wanted to split the logic even more and propose the following 3 commands at the end: `brew test`, `brew build-bottle` and `brew ci-upload`.

Of course there will be shared code amongst the 3 scripts: we will need a library/tools/utils file for that.

> I'm not sure it's worth splitting up what is essentially just a CI script. Instead of moving into multiple files what about splitting into more methods or modules?

I agree that splitting into more methods/modules is also something I would like to do. It would maybe make clearer what the different parts of the code are doing. But that does not solve the issue that the script is too long and trying to do too many different things at once.

MikeMcQuaid:
> I'm ok with that strategy too.

Unfortunately that means that it has to remain almost indefinitely to avoid breaking the workflow of taps using it.

> I feel that the `brew-test-bot.rb` file has grown too much.

It has always been a pretty long/big file.

> The script does 3 different things from a functional point of view (if I did not miss other things): test brew, build bottles, upload bottles.
> 
> This change is the first step: I wanted to split the logic even more and propose the following 3 commands at the end: `brew test`, `brew build-bottle` and `brew ci-upload`.

You can't really separate the testing and building bottle stages. The CI upload stage can be split but, as you've mentioned, it needs to share code and therefore is not really a good candidate.

> Of course there will be shared code amongst the 3 scripts: we will need a library/tools/utils file for that.

I'm not convinced that will make things easier to read or debug.

> I agree that splitting into more methods/modules is also something I would like to do. It would maybe make clearer what the different parts of the code are doing. But that does not solve the issue that the script is too long and trying to do too many different things at once.

Splitting into more methods and modules will facilitate the splitting into different files or commands if desired in future.

Large-scale refactoring of this script is _extremely_ likely to break things and some of the things that could break are around very sensitive areas. Any refactoring should be preceded by adding additional tests and testing. A better fit still would be to figure out ways to remove code from here e.g. removing support for things or flags that are no longer used.

I'm not sure where the motivation has come from for this at all, honestly? Can you elaborate? If there's issues that need to be fixed in `brew test-bot` could you file them on this repository?

iMichka:
> Unfortunately that means that it has to remain almost indefinitely to avoid breaking the workflow of taps using it.

That's because this repo has no releases/versions, and no deprecation cycle. This prevents from doing any API change here.

> It has always been a pretty long/big file.

It does not need to stay like this.

> You can't really separate the testing and building bottle stages. The CI upload stage can be split but, as you've mentioned, it needs to share code and therefore is not really a good candidate.

That's why we have includes. Shared code can be moved to a separate utils/tools file.

> I'm not convinced that will make things easier to read or debug.

It is not a silver bullet. My experience is different here, and I have worked with many teams that tried to keep their scripts small, and moved shared code to separate files. But I have also seen teams with 10000 line long monoliths. It is maybe just a different programming style and a preference. I personally favour smaller scripts when possible.

> Splitting into more methods and modules will facilitate the splitting into different files or commands if desired in future.

I totally agree that one step should also be to move some code in better understandable methods.

>Large-scale refactoring of this script is extremely likely to break things and some of the things that could break are around very sensitive areas. Any refactoring should be preceded by adding additional tests and testing. A better fit still would be to figure out ways to remove code from here e.g. removing support for things or flags that are no longer used.

I also totally agree that this code needs more tests. And I know that this code is quite fragile for having read it multiple times when looking for bugs in our linuxbrew CI.

Removing more code would also be nice. For example, `resolve_test_tap` tries to find out the tap through various env variables ... we could just pass the tap as argument to the script (and make it mandatory to pass that argument). That's almost 36 lines of code that can be deleted. And there are probably other places where the code could maybe be migrated to brew?

> I'm not sure where the motivation has come from for this at all, honestly? Can you elaborate? If there's issues that need to be fixed in brew test-bot could you file them on this repository?

I have some smaller bugs to fix for the linuxbrew CI. I am running this script sometimes locally and see errors/warnings pop up a different places. Nothing big, I have workarounds for these. It's maybe just me passing the wrong env variables.

So while I wanted to work on these, I thought: let's refactor this code before trying to solve the issue directly, so the next time I need to dig through it, it will be easier.
Also, there are not so many people that know that script very well. Last summer I got stuck for hours in this script as some stuff got broken at one point through various brew updates, and @sjackman was on holidays. I felt that I needed to be more independent and be able to fix that myself in the future, so this is also one of the reasons I am playing around in that part of the code.

I kept this PR as a WIP, there nothing is mandatory here. It is just maybe me who prefers this programming style. It's okay for me to pass on this, I can also directly debug my problems and provide improvements without changing anything to the core structure of the code. I just thought that my change would maybe be helpful for others.

MikeMcQuaid:
> That's because this repo has no releases/versions, and no deprecation cycle. This prevents from doing any API change here.

This is the same as any tap. I don't see the need for releases/versions for this tap.

> It does not need to stay like this.

Agreed but if there's problems that haven't existed for the entirety of this script then I don't think they can be attributed to "this file is long".

> That's why we have includes. Shared code can be moved to a separate utils/tools file.

It can be. Homebrew/homebrew-bundle would be a model but you'd need to write tests at the same time as doing any refactoring.

> I personally favour smaller scripts when possible.

So do I but that's somewhat irrelevant. The question is: what will be the benefits of moving this to smaller scripts vs. the trade-offs? It is almost 100% certain that bugs will be introduced as part of any refactoring. As a result, responsible refactoring relies on adding tests before or at the same time as any refactoring.

> For example, `resolve_test_tap` tries to find out the tap through various env variables ... we could just pass the tap as argument to the script (and make it mandatory to pass that argument).

Again, that would break existing taps.

> So while I wanted to work on these, I thought: let's refactor this code before trying to solve the issue directly, so the next time I need to dig through it, it will be easier.

Refactoring to make bug fixing easier is not a strong argument, I'm afraid, given that refactoring (without high unit and integration test coverage) usually introduces bugs.

> It's okay for me to pass on this, I can also directly debug my problems and provide improvements without changing anything to the core structure of the code. I just thought that my change would maybe be helpful for others.

I'm 👍 on improving the code but I don't feel that copy-pasting some of it into a new file is the right starting point for that, sorry. If this PR turns into something that adds tests/removes unused code or options/splits into more classes or files or modules then I'd be open to reviewing that.

iMichka:
> I'm 👍 on improving the code but I don't feel that copy-pasting some of it into a new file is the right starting point for that, sorry. If this PR turns into something that adds tests/removes unused code or options/splits into more classes or files or modules then I'd be open to reviewing that.

Let's try to use my energy the right way then, and let's got for a different strategy: the goal is to get the linuxbrew changes to this script merged, to be able to shut down our brew-test bot repo. I'll work on that, and while I do the changes, I'll try to see if I can make smaller refactorings here and there, or even add some tests.

MikeMcQuaid:
> Let's try to use my energy the right way then, and let's got for a different strategy: the goal is to get the linuxbrew changes to this script merged, to be able to shut down our brew-test bot repo. I'll work on that, and while I do the changes, I'll try to see if I can make smaller refactorings here and there, or even add some tests.

Sounds perfect, thanks @iMichka ❤️

PR #225, opened 2019-01-12
test-bot: don't use unlink --force.

MikeMcQuaid:
This isn't a documented flag and doesn't do anything.


fxcoudert:
Thanks a lot @MikeMcQuaid

PR #229, opened 2019-01-24
azuze-pipelines-yml: early exit on errors.

MikeMcQuaid:
We don't want to silently ignore failing commands.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,6 +4,7 @@ jobs:
     vmImage: macOS-10.13
   steps:
     - bash: |
+        set -e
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -4,6 +4,7 @@ jobs:
     vmImage: macOS-10.13
   steps:
     - bash: |
+        set -e
```
sjackman: I suggest `set -eu`. You may want to consider `set -eux`.

MikeMcQuaid: To be consistent with the other PRs will leave as-is for now.

PR #235, opened 2019-02-13
test-bot: check linkage of formula being tested.

MikeMcQuaid:
Rather than just the dependents.


sjackman:
Thanks, Mike!

iMichka:
We were discussing a similar thing on linuxbrew lately: https://github.com/Linuxbrew/homebrew-test-bot/pull/138

Though superenv should prevent any weird linkages, I guess that it does not hurt to check the linkage after the build has passed.

PR #236, opened 2019-02-13
brew-test-bot: get modified formulae for merge commits

iMichka:


iMichka:
The current check for this feature is on merge commits, but we may restrict that workflow to the linuxbrew/homebrew-core tap only

sjackman:
The motivation for this PR is to support the Linuxbrew/core workflow of merging Homebrew/core/master into Linuxbrew/core/master.

To merge Homebrew/core into Linuxbrew/core we…
1. merge Homebrew/core/master into Linuxbrew/core/master (creates a merge commit)
2. submit that merge as a PR, so the `HEAD` commit is a merge commit

Rather than rebuilding very single formula that was touched, we rebuild only those formulae whose bottle has changed.

@iMichka How many diff chunks remain between Homebrew/test-bot and Linuxbrew/test-bot? Is this the only one?

iMichka:
@sjackman yes, only that one, and the REQUIRED_TAPS code on top of the file:
```
REQUIRED_TAPS = [CoreTap.instance.name] + %w[
    linuxbrew/test-bot
    linuxbrew/xorg
  ].freeze
```

We are almost there.

Regarding this particular change: it is a little bit ugly to look at the merge commit text and parse it to get the list of changes, but it has been working well for the last years.

sjackman:
Other than saving time, the big reason for this PR is that building bottles for formulae whose bottle was not changed upstream causes the build to fail at the `brew bottle --merge --keep-old` step, because those formulae already have Linux bottles.

MikeMcQuaid:
Makes sense, thanks @iMichka!

MikeMcQuaid:
> and the REQUIRED_TAPS code on top of the file



> @sjackman yes, only that one, and the REQUIRED_TAPS code on top of the file:

Added in https://github.com/Homebrew/homebrew-test-bot/pull/237

iMichka:
Thanks @MikeMcQuaid !

sjackman:
🎉

PR #241, opened 2019-02-19
brew-test-bot: BUILD_REPOSITORY_URI may contain token, fix regex

iMichka:


iMichka:
If I print BUILD_REPOSITORY_URI on azur-pipelines (in a pull request on linuxbrew-core), the URI is printed as: ***. @sjackman guessed that there may be a secret token in the URI. I could not find any documentation on this though. Changing the regex allows the code to proceed and the Tap object to be correctly instantiated.

MikeMcQuaid:
Thanks @iMichka!

PR #253, opened 2019-05-08
Refactor: move bulk of test-bot logic to lib/test-bot.rb

scpeters:
This is a first step in refactoring that moves the bulk of the logic from `cmd/brew-test-bot.rb` into `lib/test-bot.rb`, following the pattern set by [homebrew-bundle](https://github.com/Homebrew/homebrew-bundle) as discussed in the following discourse post:

* https://discourse.brew.sh/t/brew-test-bot-refactor-cmd-brew-test-bot-rb/4758/8

In 8aba5e1, the `cmd/brew-test-bot.rb` file is copied to `lib/test-bot.rb` and then the `def test_bot`  and `Homebrew.test_bot` (the last line) are removed from the `lib` file, while everything else is removed from the `cmd` file and replaced with `require 'test-bot'`. Both files still have all the documentation in the header of the file.

In e63a4f7, `../lib` is added to the `$LOAD_PATH` so that `require 'test-bot'` actually works.

When refactoring, I try keep big moves separate from modifications, so I would appreciate suggestions on where code should be moved in this pull request or if some documentation should be dropped from the `lib` file, but I'd prefer that other modifications be saved for a follow-up.


scpeters:
If there's a way to do this while preserving `git blame` for `lib/test-bot.rb`, I would love to know it.

scpeters:
I suppose an alternative would be to move a small bit of code to `lib/test-bot.rb` or `lib/test-bot/*.rb` rather than everything to `lib/test-bot.rb` at once.

MikeMcQuaid:
> If there's a way to do this while preserving `git blame` for `lib/test-bot.rb`, I would love to know it.

There isn't unfortunately.

> lib/test-bot/*.rb

I'd rather see this happen. Could be done at least per-class e.g. `class Test` and `class Step` and everything else could live in separate files. I'd also like to see one, token unit test added per new file as I'm not sure splitting things into separate files itself makes things better (but could be convinced otherwise or at least help with the testing setup).

scpeters:
the `Step` class is perhaps the smallest bit at less than 200 lines; I'd be happy to make a separate pull request that moves that moves only that class into `lib/test-bot/step.rb` and I'd be happy to make a spec file, but I'm not sure how to set up the test infrastructure (a Rakefile?)

MikeMcQuaid:
> the `Step` class is perhaps the smallest bit at less than 200 lines; I'd be happy to make a separate pull request that moves that moves only that class into `lib/test-bot/step.rb`

Sounds good 👍 



> but I'm not sure how to set up the test infrastructure (a Rakefile?)

Just copy-paste from the Homebrew/homebrew-bundle repo and I can help trim things down/edit as necessary.

Thanks again @scpeters!

scpeters:
> the Step class is perhaps the smallest bit at less than 200 lines; I'd be happy to make a separate pull request that moves that moves only that class into lib/test-bot/step.rb

I've started a WIP PR in #259 for this class. closing this big one in favor of the smaller one first.

PR #254, opened 2019-05-10
Upgrade Azure to MacOS 10.14 and Xcode 10.2

amyspark:
This pull request upgrades Azure to use MacOS Mojave and Xcode 10.2.

/cc Homebrew/brew#6113, @MikeMcQuaid.


MikeMcQuaid:
Thanks again @amyspark!

Issue #251, opened 2019-04-24
Question - How does Jenkins know which formula to bottle?


ladislas:
Hi there! 

Following on my quest to automatize the process of building and publishing bottles for taps, I've spent quite some time studying Jenkins' console outputs and the `brew-test-bot.rb` source code.

One thing I can't get my head around is this:

> How does Jenkins know which formula to bottle?

If we take for example https://github.com/Homebrew/homebrew-core/pull/39211, the corresponding logs are:

- general log: https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/41069/console
- high sierra log: https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/41069/version=high_sierra/console

In the High Sierra log, one can find this line...

```
brew pull --clean https://github.com/Homebrew/homebrew-core/pull/39211
```

...where the pull url comes from the `ghprbPullLink` variable.
And then it magically seems to know which formula to test and bottle (`qbs`)!

```
brew uses --recursive qbs
```

How could I reproduce the same behavior in Azure Pipelines? 

For the moment I'm passing the formula name as a parameter to `brew test-bot` but what if I have multiple formulae?

Thanks a lot for your help!

Best,  
-- Ladislas

sjackman:
Here's the magic line:
https://github.com/Homebrew/homebrew-test-bot/blob/fa91172273887efb8aa799a3401cced7690cbf0f/cmd/brew-test-bot.rb#L556-L557

ladislas:
Thanks @sjackman! That was my guess :)

So doest that mean I have to pass `--tap=username/repo` when calling `brew test-bot` to bottle tap's formulae in Azure?

MikeMcQuaid:
@ladislas You shouldn't need to, no. I'm glad you're working on this as we plan to (eventually) use Azure Pipelines to bottle some/all Homebrew/homebrew-core formulae.

MikeMcQuaid:
@ladislas Closing this as the question is answered. In general can you use Discourse for questions? Feel free to mention me by username there and I'll reply quicker. Thanks again!

ladislas:
> In general can you use Discourse for questions? 

Sure! Will do!

PR #255, opened 2019-05-10
Fix detection of modified formulae on Azure

ladislas:
On Azure, the "magic" was not happening (https://github.com/Homebrew/homebrew-test-bot/issues/251#issuecomment-486345482).

This PR fixes the issue and add a comment about `git checkout $(System.PullRequest.SourceBranch)` before calling `brew test-bot`.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -413,6 +413,7 @@ module Homebrew
       elsif ENV["BUILD_REPOSITORY_URI"] && ENV["SYSTEM_PULLREQUEST_PULLREQUESTNUMBER"]
         @url = "#{ENV["BUILD_REPOSITORY_URI"]}/pull/#{ENV["SYSTEM_PULLREQUEST_PULLREQUESTNUMBER"]}"
         @hash = nil
+        system("git", "-C", @repository, "checkout", ENV["SYSTEM_PULLREQUEST_SOURCEBRANCH"])
       end
 
       # Use Jenkins Git plugin variables for master branch jobs.
@@ -433,7 +434,7 @@ module Homebrew
         diff_start_sha1 =
           Utils.popen_read("git", "-C", @repository, "rev-parse",
                                   "--short",
-                                  ENV["SYSTEM_PULLREQUEST_TARGETBRANCH"]).strip
+                                  "origin/#{ENV["SYSTEM_PULLREQUEST_TARGETBRANCH"]}").strip
         diff_end_sha1 = current_sha1
       # Use CircleCI Git variables.
       elsif ENV["CIRCLE_SHA1"]


sjackman:
> Make sure you ran `git checkout $(System.PullRequest.SourceBranch)` in a previous step or diff_start_sha1 will be empty.

`git fetch origin $(System.PullRequest.SourceBranch)` should be sufficient.

MikeMcQuaid:
> `git fetch origin $(System.PullRequest.SourceBranch)` should be sufficient.

Agreed. Reckon you can remove the comment accordingly?

ladislas:
Thanks for your feedback guys! Let me test it first and I'll update the PR.

Should we `git fetch origin` for the users (and remove the comment) or should the users do it themselves in their `azure-pipelines.yml` (and keep the comment?)?

MikeMcQuaid:
Do it for them 👍

ladislas:
@MikeMcQuaid @sjackman after testing, and if I'm not doing anything wrong, I have the following results:

This works ✅:

```ruby
system("git", "-C", @repository, "checkout", ENV["SYSTEM_PULLREQUEST_SOURCEBRANCH"])
diff_start_sha1 =
          Utils.popen_read("git", "-C", @repository, "rev-parse",
                                  "--short",
                                  "origin/#{ENV["SYSTEM_PULLREQUEST_TARGETBRANCH"]}").strip
```

This does **not** work ❌:

```ruby
system("git", "-C", @repository, "fetch", "origin", ENV["SYSTEM_PULLREQUEST_SOURCEBRANCH"])
diff_start_sha1 =
          Utils.popen_read("git", "-C", @repository, "rev-parse",
                                  "--short",
                                  "origin/#{ENV["SYSTEM_PULLREQUEST_TARGETBRANCH"]}").strip
```

MikeMcQuaid:
> ```ruby
> system("git", "-C", @repository, "fetch", "origin", ENV["SYSTEM_PULLREQUEST_SOURCEBRANCH"])
> ```

`git fetch` doesn't take a branch argument, just a remote.

ladislas:
@MikeMcQuaid I just tried with the following and it does not work either:

```ruby
system("git", "-C", @repository, "fetch", "origin")
```

In Azure, the full git history seems to be present, but they put you on a merge between the target branch and the source branch.

MikeMcQuaid:
> ```ruby
> system("git", "-C", @repository, "fetch", "origin")
> ```

That will fetch the latest `origin` branches.

> In Azure, the full git history seems to be present, but they put you on a merge between the target branch and the source branch.

That's fine. You can use e.g. `git merge-base origin/master HEAD` to use the merge base as the `start_sha`. This is essentially the same algorithm that pull requests use to decide what to show as a diff.

ladislas:
Thanks @MikeMcQuaid! But I must admit I'm a bit lost now :(

How should I update the PR?

MikeMcQuaid:
@ladislas Me too 😉. Can you explain what the current problem is, how you're trying to fix it and what doesn't work with the versions @sjackman and I have suggested?

ladislas:
@MikeMcQuaid Sure! :) 

My issue is that on Azure Pipelines, the following doesn't work and returns nothing - https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L433

```ruby
diff_start_sha1 =
  Utils.popen_read("git", "-C", @repository, "rev-parse",
                          "--short",
                          ENV["SYSTEM_PULLREQUEST_TARGETBRANCH"]).strip
```

After hours of investigating, adding `puts` and so on (which is one of the reason for the proposal to refactor the code), I found out that the following fixed the issue: 

- adding `origin/` in front of `ENV["SYSTEM_PULLREQUEST_TARGETBRANCH"]`
- and calling `git checkout ENV["SYSTEM_PULLREQUEST_SOURCEBRANCH"]` **brefore** calling `brew test-bot --root-url`, which I initially did in my `azure-pipelines.yml` but then moved to `brew-test-bot.rb`, which gives us the following that works and fixes my issue without adding to much complexity:

```ruby
system("git", "-C", @repository, "checkout", ENV["SYSTEM_PULLREQUEST_SOURCEBRANCH"])
diff_start_sha1 =
  Utils.popen_read("git", "-C", @repository, "rev-parse",
                          "--short",
                          "origin/#{ENV["SYSTEM_PULLREQUEST_TARGETBRANCH"]}").strip

```

I've tried your suggestions and all the combinations but nothing seems to work.

I'm happy to try something else if you thing that if might work better. The `git merge-base origin/master HEAD` gives me a commit but it fails later (hard to tell where though but I can investigate if needed).

That being said, IMHO, the solution I came up with is *super basic* and does not introduce much changes. Also it only targets Azure Pipelines so the side effects (if any but I doubt it) are constrained and limited to this CI platform.

Let me know what you thing! :)

ladislas:
I've updated the PR based on my previous explanation.

cmd/brew-test-bot.rb
```diff
@@ -430,10 +430,11 @@ def download
         diff_end_sha1 = current_sha1
       # Use Azure Pipeline variables for master or branch jobs.
       elsif ENV["SYSTEM_PULLREQUEST_TARGETBRANCH"] && ENV["BUILD_SOURCEVERSION"]
+        system("git", "-C", @repository, "checkout", ENV["SYSTEM_PULLREQUEST_SOURCEBRANCH"])
```
MikeMcQuaid: Can this turn into `test "git", "-C", @repository, "checkout", ....` up at line 416 to match something similar with Jenkins on line 399?

MikeMcQuaid: (Once that's done I think this is good to merge)

ladislas: Done and tested! It works 👍

MikeMcQuaid:
Thanks again @ladislas, great work here 👏.

ladislas:
Thanks to you for being open and responsive to contributions :)

PR #256, opened 2019-05-13
Revert "Fix detection of modified formulae on Azure"

MikeMcQuaid:
Reverts Homebrew/homebrew-test-bot#255

PR #258, opened 2019-05-13
add ohai to help debug start and end diff sha1

ladislas:
as discussed in #256

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -538,11 +538,11 @@ module Homebrew
                  "log", "-1", "--format=%h (%s)"
         ).strip
         puts <<~EOS
-          Testing tap #{@tap} with:
-            - origin/master   - #{tap_origin_master_revision.blank? ? "(undefined)" : tap_origin_master_revision}
-            - HEAD            - #{tap_revision.blank? ? "(undefined)" : tap_revision}
-            - diff_start_sha1 - #{diff_start_sha1.blank? ? "(undefined)" : diff_start_sha1}
-            - diff_end_sha1   - #{diff_end_sha1.blank? ? "(undefined)" : diff_end_sha1}
+          \nTesting tap #{@tap} with:
+            origin/master   #{tap_origin_master_revision.blank? ? "(undefined)" : tap_origin_master_revision}
+            HEAD            #{tap_revision.blank? ? "(undefined)" : tap_revision}
+            diff_start_sha1 #{diff_start_sha1.blank? ? "(undefined)" : diff_start_sha1}
+            diff_end_sha1   #{diff_end_sha1.blank? ? "(undefined)" : diff_end_sha1}
         EOS
       end
 
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -537,15 +537,16 @@ module Homebrew
           "git", "-C", @tap.path.to_s,
                  "log", "-1", "--format=%h (%s)"
         ).strip
-        puts <<~EOS
-
-          Testing tap #{@tap} with:
-            origin/master   #{tap_origin_master_revision.blank? ? "(undefined)" : tap_origin_master_revision}
-            HEAD            #{tap_revision.blank? ? "(undefined)" : tap_revision}
-            diff_start_sha1 #{diff_start_sha1.blank? ? "(undefined)" : diff_start_sha1}
-            diff_end_sha1   #{diff_end_sha1.blank? ? "(undefined)" : diff_end_sha1}
-        EOS
       end
+
+      puts <<~EOS
+
+        Testing#{" tap #{@tap}" if @tap.present?} with:
+          origin/master   #{tap_origin_master_revision.blank? ? "(undefined)" : tap_origin_master_revision}
+          HEAD            #{tap_revision.blank? ? "(undefined)" : tap_revision}
+          diff_start_sha1 #{diff_start_sha1.blank? ? "(undefined)" : diff_start_sha1}
+          diff_end_sha1   #{diff_end_sha1.blank? ? "(undefined)" : diff_end_sha1}
+      EOS
 
       return if diff_start_sha1 == diff_end_sha1
       return if @url && steps.last && !steps.last.passed?
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -533,12 +533,17 @@ module Homebrew
           "git", "-C", @tap.path.to_s,
                  "log", "-1", "--format=%h (%s)", "origin/master"
         ).strip
-        puts "#{@tap} origin/master #{tap_origin_master_revision}"
         tap_revision = Utils.popen_read(
           "git", "-C", @tap.path.to_s,
                  "log", "-1", "--format=%h (%s)"
         ).strip
-        puts "#{@tap} HEAD #{tap_revision}"
+        puts <<~EOS
+          Testing tap #{@tap} with:
+            - origin/master   - #{tap_origin_master_revision.blank? ? "(undefined)" : tap_origin_master_revision}
+            - HEAD            - #{tap_revision.blank? ? "(undefined)" : tap_revision}
+            - diff_start_sha1 - #{diff_start_sha1.blank? ? "(undefined)" : diff_start_sha1}
+            - diff_end_sha1   - #{diff_end_sha1.blank? ? "(undefined)" : diff_end_sha1}
+        EOS
       end
 
       return if diff_start_sha1 == diff_end_sha1
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -538,7 +538,8 @@ module Homebrew
                  "log", "-1", "--format=%h (%s)"
         ).strip
         puts <<~EOS
-          \nTesting tap #{@tap} with:
+
+          Testing tap #{@tap} with:
             origin/master   #{tap_origin_master_revision.blank? ? "(undefined)" : tap_origin_master_revision}
             HEAD            #{tap_revision.blank? ? "(undefined)" : tap_revision}
             diff_start_sha1 #{diff_start_sha1.blank? ? "(undefined)" : diff_start_sha1}


cmd/brew-test-bot.rb
```diff
@@ -446,6 +446,8 @@ def download
         diff_end_sha1 = diff_start_sha1 = current_sha1
       end
 
+      ohai "Testing commits #{diff_start_sha1} to #{diff_end_sha1}"
```
scpeters: should this go after the next statement, which sets `diff_start_sha1`?

scpeters: the macOS build prints `==> Testing commits  to 6cb92c0`

scpeters:
I don't see the ohai statement in the Linux build

cmd/brew-test-bot.rb
iMichka: We could put a second statement just before the `return if diff_start_sha1 == diff_end_sha1` line. As `diff_start_sha1` may be modified at least twice after this line. So we would get the initial value and the modified value.

ladislas: > the macOS build prints `==> Testing commits to 6cb92c0`

Well this could be because the start sha1 is not computed on Azure :\

ladislas:
@scpeters @iMichka I've added your suggestions.

cmd/brew-test-bot.rb
```diff
@@ -446,10 +446,14 @@ def download
         diff_end_sha1 = diff_start_sha1 = current_sha1
       end
 
+      ohai "Testing commits start #{diff_start_sha1} to end #{diff_end_sha1}"
+
       diff_start_sha1 =
         Utils.popen_read("git", "-C", @repository, "merge-base",
                                 diff_start_sha1, diff_end_sha1).strip
 
+      ohai "Testing commits start #{diff_start_sha1} to end #{diff_end_sha1}"
```
sjackman:
```suggestion
      ohai "Considering testing commits from ancestor #{diff_start_sha1} to #{diff_end_sha1}"
```

cmd/brew-test-bot.rb
```diff
@@ -541,6 +545,8 @@ def download
         puts "#{@tap} HEAD #{tap_revision}"
       end
 
+      ohai "Testing commits start #{diff_start_sha1} to end #{diff_end_sha1}"
```
sjackman:
```suggestion
      ohai "Testing commits #{diff_start_sha1} to #{diff_end_sha1}"
```

cmd/brew-test-bot.rb
```diff
@@ -446,10 +446,14 @@ def download
         diff_end_sha1 = diff_start_sha1 = current_sha1
       end
 
+      ohai "Testing commits start #{diff_start_sha1} to end #{diff_end_sha1}"
```
sjackman:
```suggestion
      ohai "Consider testing commits #{diff_start_sha1} to #{diff_end_sha1}"
```

cmd/brew-test-bot.rb
sjackman: The question then is why is `diff_start_sha1` not computed on Azure?

sjackman:
I'd like to either
print a different message each time. It could create confusion to print the same message three times.
or print the message just once at the most appropriate time.

iMichka:
I would remove the print in the middle, and keep only two:
```
1) Considering testing commits #{diff_start_sha1} to #{diff_end_sha1}
2) Testing commits #{diff_start_sha1} to #{diff_end_sha1}
```

It's useful to have the initial value, and then the final one.

Ideally we could also print something each time the start/end value is modified, to know what part of the code is doing that. But the code is not modular enough right now to do this.

sjackman:
I'm fine with Michka's suggestion.

ladislas:
@sjackman - I've just pushed the changes asked by @iMichka

cmd/brew-test-bot.rb
```diff
@@ -446,6 +446,8 @@ def download
         diff_end_sha1 = diff_start_sha1 = current_sha1
       end
 
+      ohai "Considering testing commits #{diff_start_sha1} to #{diff_end_sha1}"
```
MikeMcQuaid: This looks weird if it is empty: https://dev.azure.com/Homebrew/Homebrew/_build/results?buildId=9315&view=logs&jobId=a5e52b91-c83f-5429-4a68-c246fc63a4f7&taskId=00f564b5-c6ae-5329-43c9-96bc79a72ec9&lineStart=33&lineEnd=33&colStart=1&colEnd=44

cmd/brew-test-bot.rb
```diff
@@ -541,6 +543,8 @@ def download
         puts "#{@tap} HEAD #{tap_revision}"
       end
 
+      ohai "Testing commits #{diff_start_sha1} to #{diff_end_sha1}"
```
MikeMcQuaid: This looks weird if it is empty: https://dev.azure.com/Homebrew/Homebrew/_build/results?buildId=9315&view=logs&jobId=a5e52b91-c83f-5429-4a68-c246fc63a4f7&taskId=00f564b5-c6ae-5329-43c9-96bc79a72ec9&lineStart=45&lineEnd=45&colStart=1&colEnd=32

cmd/brew-test-bot.rb
```diff
@@ -541,6 +543,8 @@ def download
         puts "#{@tap} HEAD #{tap_revision}"
```
MikeMcQuaid: This (and line 538) are already outputting the revision of the tap. I'd suggest these messages be combined or adjusted instead of duplicated.

cmd/brew-test-bot.rb
MikeMcQuaid: I don't think `ohai` makes sense for these as they are otherwise only used for commands and it looks messy/confusing to have these be the only non-command values.

ladislas: @MikeMcQuaid is there something like `Debug`? that we could turn on/off when needed? Or would a simple `puts` make the job?

cmd/brew-test-bot.rb
ladislas: @MikeMcQuaid same answer as https://github.com/Homebrew/homebrew-test-bot/pull/258#discussion_r283670972

cmd/brew-test-bot.rb
MikeMcQuaid: `puts` is fine but let's figure out a way to combine it with https://github.com/Homebrew/homebrew-test-bot/pull/258#discussion_r283655711

cmd/brew-test-bot.rb
sjackman: I believe Mike is suggesting printing something else more helpful when `diff_start_sha1` or `diff_end_sha1` is undefined, like `Testing commits (undefined) to abcdef0` rather than the current `Testing commits to abcdef0`

sjackman:
> As @scpeters mentioned: I also don't see these values on the Linux CI job. This is probably because the environment variables are not passed through to Docker. @sjackman may be able to provide advice on that.

Brewsci/bio is not using Docker and instead using Azure `vmImage: ubuntu-16.04`. That's one option for a third-party tap.

Your test tap does not appear to be using Docker either:
https://github.com/ladislas/homebrew-hello-world/blob/master/azure-pipelines.yml

In commit e970dbb45db58db2243db9f8c30a3e8647c882a9, Homebrew/test-bot was modified to use Docker. It appears to also unintentionally have started testing `homebrew/homebrew-test-bot origin/master` rather than the current PR. The code of the current PR is not being passed on through to the Docker container.

sjackman:
> The code of the current PR is not being passed on through to the Docker container.

The two options to get `test-bot` to test the current PR rather than `origin/master` is to either
1. pass the code through to Docker in a volume
2. pass sufficient environment variables through to Docker to get it to check out the right code to test

Here's the list of environment variables that are currently being passed through to Docker:

https://github.com/Homebrew/brew/blob/191898ced54a33a39e05552fe598d5a530754042/Dockerfile.yml#L11

```
HOME=/home/linuxbrew
BUILD_REASON
BUILD_REPOSITORY_URI
BUILD_SOURCEVERSION
HOMEBREW_GITHUB_API_TOKEN
SYSTEM_PULLREQUEST_PULLREQUESTNUMBER
SYSTEM_PULLREQUEST_TARGETBRANCH
TF_BUILD
```

Note that when I asked

cmd/brew-test-bot.rb
MikeMcQuaid: Yep or just ensuring the values are always set. I would also like to see (as mentioned below) this consolidated or replacing the existing printed values.

MikeMcQuaid:
> It appears to also unintentionally have started testing `homebrew/homebrew-test-bot origin/master` rather than the current PR. The code of the current PR is not being passed on through to the Docker container.

This is likely while the revert had to happen; the correct code wasn't being tested.

For this repo specifically we could consider adding a third job which tests on test-bot on Linux outside of Docker. I do think we ideally want all three ways (macOS, Linux outside Docker, Linux insider Docker) to work, though.

> 2\. pass sufficient environment variables through to Docker to get it to check out the right code to test

This seems preferable to me 👍

ladislas:
> Your test tap does not appear to be using Docker either:

@sjackman no you're right, I haven't tried Linux yet, it's on my todo once macOS bottles are up and running.

@MikeMcQuaid @sjackman - regarding the docker thingy, I have absolutely zero knowledge of that, so it should (if needed) be addressed in a separate PR/discussion(?)

Circling back to logging the start and end sha1:

- I can make the messages better
- regarding the `if @tap`, I'm not sure how to combine it with the rest as as `@tap` might not be set (or is it always set?)
- @MikeMcQuaid regarding ensuring the values are always set: do you mean the Azure Pipelines values? or the diff sha1?

@sjackman's solution here (https://github.com/Homebrew/homebrew-test-bot/pull/256#issuecomment-491990986) might be a good solution, but we would need to test it for real.

MikeMcQuaid:
> @MikeMcQuaid @sjackman - regarding the docker thingy, I have absolutely zero knowledge of that, so it should (if needed) be addressed in a separate PR/discussion(?)

We'll need to address it here to ensure that any changes made here work on the Docker CI job.

> * regarding the `if @tap`, I'm not sure how to combine it with the rest as as `@tap` might not be set (or is it always set?)

Well, I guess my question is: does that not communicate any of the information you already need to know? What information does adding the start/end provide that printing the HEAD and origin/master does not?

ladislas:
> Well, I guess my question is: does that not communicate any of the information you already need to know? What information does adding the start/end provide that printing the HEAD and origin/master does not?

Isn't it the other way around? As `@tap` won't always be defined thus ignoring the `if @tap` block, what does printing it when defined adds to printing the HEAD and origin/master that we might print all the time? Or should we not print anything if not `@tap`?

MikeMcQuaid:
If there's no tap it's testing Homebrew/brew.

ladislas:
> If there's no tap it's testing Homebrew/brew.

That explains everything then! :) Sorry for all my questions! I'll update the PR!

MikeMcQuaid:
> Sorry for all my questions!

No need to apologise, I'm really happy to have you working on this and would love to answer as many questions as I can!

sjackman:
>> @MikeMcQuaid @sjackman - regarding the docker thingy, I have absolutely zero knowledge of that, so it should (if needed) be addressed in a separate PR/discussion(?)

> We'll need to address it here to ensure that any changes made here work on the Docker CI job.

I can tackle this.

sjackman:
I've opened PR https://github.com/Homebrew/homebrew-test-bot/pull/260.

>> 1. pass the code through to Docker in a volume
>> 2. pass sufficient environment variables through to Docker to get it to check out the right code to test

> This seems preferable to me 👍

Approach 2 could work fine for other taps, but for Homebrew/test-bot itself, having `brew test-bot` update its own code to test itself seems like it could be confusing. I've gone with approach 1.

ladislas:
I've just updated the PR, you can see how it looks here:

https://dev.azure.com/ladisonline/homebrew-hello-world/_build/results?buildId=319&view=logs&jobId=f720af0a-4aeb-5402-58ab-558049540efe&taskId=b37980cf-3630-5342-dd54-7700c0688f56&lineStart=40&lineEnd=43&colStart=1&colEnd=1

Please note that it works because I'm using:

```
set -ex
if [[ -n "$SYSTEM_PULLREQUEST_PULLREQUESTNUMBER" ]]; then
  git fetch origin "master:master" "pull/$SYSTEM_PULLREQUEST_PULLREQUESTNUMBER/head:pr"
  git checkout pr
fi
```
in my `azure-pipelines.yml` (https://github.com/ladislas/homebrew-hello-world/commit/e2cec93bbb3b8e4b37b7a0182ca572f89893ba4d#diff-27eb251ec4c789653f8931379114064bR9)

ladislas:
> Please note that it works because I'm using ... in my azure-pipelines.yml

Without it, I have the following output and brew test-bot can't find the modified formulae:

```
Testing tap ladislas/hello-world with:
  - origin/master     - 8718ae8 (remove fetch origin)
  - pull request HEAD - 79eaa63 (Merge 87a2c269bc6e92861b67eb4d0c86cd806d7b6d5d into 8718ae8041bad657ffbf17bf6c7bc27e8f2739e1)
```

cmd/brew-test-bot.rb
```diff
@@ -533,12 +533,17 @@ def download
...
           "git", "-C", @tap.path.to_s,
                  "log", "-1", "--format=%h (%s)"
         ).strip
-        puts "#{@tap} HEAD #{tap_revision}"
+        puts "\nTesting tap #{@tap} with:"
+        puts "  - origin/master     - #{tap_origin_master_revision}"
+        puts "  - pull request HEAD - #{tap_revision}"
+      else
+        puts "\nTesting Homebrew/brew #{@tap} with commits:"
+        puts "    - diff_start_sha1 #{diff_start_sha1.empty? ? "(undefined)" : diff_start_sha1}"
```
MikeMcQuaid: Use `.blank?` here.

cmd/brew-test-bot.rb
```diff
@@ -533,12 +533,17 @@ def download
           "git", "-C", @tap.path.to_s,
                  "log", "-1", "--format=%h (%s)", "origin/master"
         ).strip
-        puts "#{@tap} origin/master #{tap_origin_master_revision}"
         tap_revision = Utils.popen_read(
           "git", "-C", @tap.path.to_s,
                  "log", "-1", "--format=%h (%s)"
         ).strip
-        puts "#{@tap} HEAD #{tap_revision}"
+        puts "\nTesting tap #{@tap} with:"
```
MikeMcQuaid: Use `<<~EOS` which allows you to embed newlines and print multiple lines.

cmd/brew-test-bot.rb
```diff
@@ -533,12 +533,17 @@ def download
...
+        puts "  - origin/master     - #{tap_origin_master_revision}"
+        puts "  - pull request HEAD - #{tap_revision}"
```
MikeMcQuaid: This may not be a pull request.

cmd/brew-test-bot.rb
```diff
@@ -533,12 +533,17 @@ def download
...
+      else
+        puts "\nTesting Homebrew/brew #{@tap} with commits:"
```
MikeMcQuaid: `puts "Homebrew/brew #{brew_version} (#{brew_commit_subject})"` covers some of this above.

cmd/brew-test-bot.rb
MikeMcQuaid: These can still be printed in addition to the other messages if they are useful and not always the same.

cmd/brew-test-bot.rb
ladislas: Of course! Completely forgot about that :)

cmd/brew-test-bot.rb
ladislas: I've updated the output

cmd/brew-test-bot.rb
ladislas: So I've removed the `else`

cmd/brew-test-bot.rb
```diff
@@ -533,12 +533,17 @@ def download
...
           "git", "-C", @tap.path.to_s,
                  "log", "-1", "--format=%h (%s)"
         ).strip
-        puts "#{@tap} HEAD #{tap_revision}"
+        puts <<~EOS
+          Testing tap #{@tap} with:
+            - origin/master   - #{tap_origin_master_revision.blank? ? "(undefined)" : tap_origin_master_revision}
+            - HEAD            - #{tap_revision.blank? ? "(undefined)" : tap_revision}
+            - diff_start_sha1 - #{diff_start_sha1.blank? ? "(undefined)" : diff_start_sha1}
+            - diff_end_sha1   - #{diff_end_sha1.blank? ? "(undefined)" : diff_end_sha1}
```
sjackman: Minor nit: I'm inclined to remove all the eight hyphens. Otherwise looks great! Thank you for this work, Ladislas!

ladislas: Of course! Do I keep the indentation and alignment?

ladislas: @sjackman and also a new line before `Testing tap` to make it clearer in the log?

cmd/brew-test-bot.rb
```diff
@@ -533,12 +533,17 @@ def download
...
                  "log", "-1", "--format=%h (%s)", "origin/master"
         ).strip
-        puts "#{@tap} origin/master #{tap_origin_master_revision}"
         tap_revision = Utils.popen_read(
           "git", "-C", @tap.path.to_s,
                  "log", "-1", "--format=%h (%s)"
         ).strip
-        puts "#{@tap} HEAD #{tap_revision}"
+        puts <<~EOS
+          \nTesting tap #{@tap} with:
```
MikeMcQuaid: Replace `\n` with an actual newline.

cmd/brew-test-bot.rb
sjackman: > Do I keep the indentation and alignment?

Yes keep the indentation and alignment, and a new line before `Testing tap` sounds good.

cmd/brew-test-bot.rb
MikeMcQuaid: Handle if `@tap` isn't set.

ladislas: You mean `nil` or `blank`? If `nil`, won't it be taken care of by the `if @tap`?

ladislas: @MikeMcQuaid or do you mean with an `else` clause?

MikeMcQuaid: Sorry, I missed that it was indented in there. I think ideally there'd be consistent output of this stuff (e.g. `diff_start_sha1`) regardless of whether a `@tap` is set.

How about pulling it out the `if` and doing something like `"Testing#{" tap #{@tap}" if tap.present?} with:"`?

After that I think we're good to go here 👍

ladislas: sounds good, I'll update the PR now :)

cmd/brew-test-bot.rb
```diff
@@ -537,16 +537,17 @@ def download
...
-
-          Testing tap #{@tap} with:
-            origin/master   #{tap_origin_master_revision.blank? ? "(undefined)" : tap_origin_master_revision}
-            HEAD            #{tap_revision.blank? ? "(undefined)" : tap_revision}
-            diff_start_sha1 #{diff_start_sha1.blank? ? "(undefined)" : diff_start_sha1}
-            diff_end_sha1   #{diff_end_sha1.blank? ? "(undefined)" : diff_end_sha1}
-        EOS
       end
 
+      puts <<~EOS
```
ladislas: @MikeMcQuaid does that work for you? :)

MikeMcQuaid: Yup, perfect!

ladislas: 👍

MikeMcQuaid:
Looks good, nice work again @ladislas!

Getting slightly different results on Mac and Linux:
- https://dev.azure.com/Homebrew/Homebrew/_build/results?buildId=9547&view=logs&jobId=a5e52b91-c83f-5429-4a68-c246fc63a4f7&taskId=00f564b5-c6ae-5329-43c9-96bc79a72ec9&lineStart=42&lineEnd=47&colStart=1&colEnd=1
- https://dev.azure.com/Homebrew/Homebrew/_build/results?buildId=9547&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=45b31e99-b18c-506f-46cf-7d3a4d5140a9&lineStart=27&lineEnd=32&colStart=1&colEnd=1

I guess that's Docker vs. non Docker @sjackman? Would be good if we got those values identical in both places; maybe https://github.com/Homebrew/homebrew-test-bot/pull/261 is part of that?

ladislas:
> Getting slightly different results on Mac and Linux:

@MikeMcQuaid Yes I think it's the same issue. `diff_start_sha1` is also undefined on macOS for the PR but not on `master`.

sjackman:
Thanks for this PR, @ladislas. It's already proving useful to highlight idiosyncrasies!

PR #259, opened 2019-05-15
WIP move Step class to lib/test-bot/step.rb

scpeters:
I'm working on adding a test. I just wanted to run it through CI to make sure it was ok so far.

I removed most of the require statements from `lib/test-bot/step.rb`, but I'm not sure if I removed too many. Also, I'm not sure if `step.rb` should have `module_function` set.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -101,6 +101,13 @@ require "development_tools"
 require "utils/bottles"
 require "json"
 
+TEST_BOT_ROOT = File.expand_path "#{File.dirname(__FILE__)}/.."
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
> I'm working on adding a test.

Cool, will review once that's the case.

> Also, I'm not sure if `step.rb` should have `module_function` set.

It shouldn't as it is an instantiated class.

MikeMcQuaid:
@scpeters Any updates on this?

scpeters:
I haven't been working on a test. It can be closed until I have some more time to spend on getting one set up.

PR #262, opened 2019-05-21
puts modified and added formulae

ladislas:
kind of like #258, this PR outputs the added and modified formulae to help debug.

I could also add `deleted_formulae` if needed.

Let me know what you think.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -578,9 +578,9 @@ module Homebrew
       puts <<~EOS
 
         Formula changes to be tested:
-          added formulae    #{@added_formulae.blank? ? "(empty)" : @added_formulae}
-          modified formulae #{@modified_formulae.blank? ? "(empty)" : @modified_formulae}
-          deleted formulae  #{@deleted_formulae.blank? ? "(empty)" : @deleted_formulae}
+          added formulae    #{@added_formulae.blank? ? "(empty)" : @added_formulae.reject(&:blank?).join(' ')}
+          modified formulae #{@modified_formulae.blank? ? "(empty)" : @modified_formulae.reject(&:blank?).join(' ')}
+          deleted formulae  #{@deleted_formulae.blank? ? "(empty)" : @deleted_formulae.reject(&:blank?).join(' ')}
       EOS
     end
 
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -578,9 +578,9 @@ module Homebrew
       puts <<~EOS
 
         Formula changes to be tested:
-          added formulae    #{@added_formulae.blank? ? "(empty)" : @added_formulae.reject(&:blank?).join(' ')}
-          modified formulae #{@modified_formulae.blank? ? "(empty)" : @modified_formulae.reject(&:blank?).join(' ')}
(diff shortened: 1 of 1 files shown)


sjackman:
I scanned the code quickly, and I didn't see that the value of `@formulae @added_formulae @modified_formulae` are printed anywhere. If that's the case, :+1: from me.

iMichka:
I like too. Nice addition.

Can we add the deleted formulae too?

cmd/brew-test-bot.rb
```diff
@@ -574,6 +574,13 @@ def download
       @formulae += @added_formulae + @modified_formulae
       @test_brew = (!@tap || @test_bot_tap) &&
                    (@formulae.empty? || @test_default_formula)
+
+      puts <<~EOS
+
+        Bottles to be built for:
+          added formulae    #{@added_formulae.empty? ? "(empty)" : @added_formulae}
```
MikeMcQuaid: Use `blank?` rather than `empty?`

cmd/brew-test-bot.rb
```diff
@@ -574,6 +574,13 @@ def download
       @formulae += @added_formulae + @modified_formulae
       @test_brew = (!@tap || @test_bot_tap) &&
                    (@formulae.empty? || @test_default_formula)
+
+      puts <<~EOS
+
+        Bottles to be built for:
```
MikeMcQuaid: Bottles aren't necessarily going to be built. How about `Formula changes to be tested:`

cmd/brew-test-bot.rb
ladislas: sure!

cmd/brew-test-bot.rb
ladislas: Sounds good to me, I'll change that.

ladislas:
I just updated the PR based on your feedback.

I've added `deleted_formulae`.

it's gonna print as `["formula1", "formula2"]` -- should it be printed as `formula1, formula2` instead?

sjackman:
> it's gonna print as ["formula1", "formula2"] -- should it be printed as formula1, formula2 instead?

Yes, separated with either comma space `, ` or simply space ` `

ladislas:
@sjackman should I use:

```ruby
@modified_formulae.join(' ') # => "formula1 formula2"

#or 

@modified_formulae.reject(&:blank?).join(' ') # => same but will remove '' and nil, just in case
```

ladislas:
I went with the safer `reject`, to handle all cases.

sjackman:
There will not be blank formulae names, so `@modified_formulae.join(' ')` is fine.

MikeMcQuaid:
> There will not be blank formulae names, so `@modified_formulae.join(' ')` is fine.

Agreed. Once this is done this should be good to go!

ladislas:
Done! ✅ (from my phone for the first time)

sjackman:
> Done! ✅ (from my phone for the first time)

Hah. Well done! Thanks again for this work, Ladislas! 🌟

MikeMcQuaid:
Thanks again @ladislas!

MikeMcQuaid:
@ladislas Please check your email 😂

PR #263, opened 2019-05-23
.github/: defer to Homebrew/.github repo.

MikeMcQuaid:
This allows maintaining more settings in one place.


sjackman:
Nice! Where is the delegation specified? I took a quick look, but didn't find it.

MikeMcQuaid:
@sjackman There's no specific setup it's just inferred from GitHub.

sjackman:
Ah, I see. Nice! https://github.com/Homebrew/.github

PR #264, opened 2019-06-11
test-bot: abort more core operations.

MikeMcQuaid:
If there's an ongoing (or, more likely, failed) `rebase` or `am` in Homebrew/homebrew-core then make sure it's cleaned up accordingly.


scpeters:
it still seems to be broken even after merging this:

* https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/43499/

PR #267, opened 2019-06-17
test-bot: don't uninstall installed formula.

### LINKED ISSUE #266: Bottles can't be found when building multiple bottles with common indirect dependents
In https://github.com/Homebrew/homebrew-core/pull/39225, both `python` and `python@2` are updated to use a new version of `setuptools`. `python` gets built first and all its dependencies are tested without any failures. Then it is uninstalled, the `python@2` bottle is built (successfully), and the dependencies of `python@2` are tested. The first formula to fail is `alot`, which depends directly on `python@2`, but not directly on `python`:

* https://github.com/Homebrew/homebrew-core/blob/master/Formula/alot.rb#L18-L25

`alot` has multiple direct dependencies that depend indirectly on `python`, such as `gpgme` and `notmuch`. For some reason, `test-bot` ends up trying to find the new python bottle on bintray instead of installing the recently built and unpublished one from the filesystem.

~~~
==> Installing alot dependency: python
==> Downloading https://homebrew.bintray.com/bottles/python-3.7.3.mojave.bottle.1.tar.gz
curl: (22) The requested URL returned error: 404 Not Found
Error: Failed to download resource "python"
Download failed: https://homebrew.bintray.com/bottles/python-3.7.3.mojave.bottle.1.tar.gz
~~~

Note that `alot` was also tested successfully when `python`'s dependencies were tested, but failed later when `python@2` dependencies were tested.

I think a workaround is to split up changes into separate pull requests, though this may not always be feasible or desired. I have observed this issue at times while maintaining the osrf/simulation tap, but I haven't reported it until now because I've found it difficult to describe succinctly with core formulae. I've observed it when building multiple bottles in a pull request and some downstream formulae depend on one of the formula directly, and the other indirectly.

I've saved the failed jenkins build log from https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/43640/version=mojave/ in the following gist:

* https://gist.github.com/scpeters/7beb8914c63b766849106a95f4ddc7c3

MikeMcQuaid:
Otherwise they may attempt to pour from bottles and fail.

Fixes #266.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -1012,10 +1012,6 @@ module Homebrew
         end
       end
 
-      unless @all_formulae_dependencies.include?(formula_name)
-        test "brew", "uninstall", "--force", formula_name
-      end
-
       return if @unchanged_dependencies.empty?
 
       test "brew", "uninstall", "--force", *@unchanged_dependencies
@@ -1269,7 +1265,6 @@ module Homebrew
 
     def formulae
       changed_formulae_dependents = {}
-      @all_formulae_dependencies = []
 
       @formulae.each do |formula|
         begin
@@ -1289,8 +1284,6 @@ module Homebrew
           safe_system "brew", "tap", e.tap.name
           retry
         end
-
-        @all_formulae_dependencies += formula_dependencies
 
         unchanged_dependencies = formula_dependencies - @formulae
         changed_dependencies = formula_dependencies - unchanged_dependencies


scpeters:
Testing now

scpeters:
this doesn't fix Homebrew/homebrew-core#39225 in my testing

commenting inline

cmd/brew-test-bot.rb
```diff
@@ -1300,6 +1300,8 @@ def formulae
         end
       end
 
+      @all_formulae_unchanged_dependencies -= @formulae
```
scpeters: I think this logic would have to be reversed to have an effect on https://github.com/Homebrew/homebrew-core/pull/39225  (`-=` -> `+=`)

which would be equivalent to removing the `brew uninstall` lines at https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L1015-L1017

there may be a more surgical approach, but I haven't had time to figure out what it would be

MikeMcQuaid: @scpeters Whoops, typo! Fixed that, mind retesting? Thanks again!

scpeters:
This PR fixes the problems with Homebrew/homebrew-core#39225, but I think there is a simpler way to implement it. I believe the following is functionally equivalent to this PR: https://github.com/scpeters/homebrew-test-bot/commit/5e5e0d708baedada01e8a3dfc928a5a3c4e8a7c7

According to the current logic, I think the `brew uninstall` command would never be executed, so I think it is clearer to just remove it.

Are we losing something by removing that command? I imagine there is a possibility for more opportunistic linkage. I'll keep thinking about it.

MikeMcQuaid:
> This PR fixes the problems with [Homebrew/homebrew-core#39225](https://github.com/Homebrew/homebrew-core/pull/39225), but I think there is a simpler way to implement it. I believe the following is functionally equivalent to this PR: [scpeters@5e5e0d7](https://github.com/scpeters/homebrew-test-bot/commit/5e5e0d708baedada01e8a3dfc928a5a3c4e8a7c7)

Yes, you're right. I'm happy to just see it removed.

> Are we losing something by removing that command? I imagine there is a possibility for more opportunistic linkage.

Yeh, there is. Ideally we'd probably calculate the reverse dependencies for all the reverse dependencies we will install and test (rather than just the `@formulae`) but that may be annoyingly slow. Thoughts?

MikeMcQuaid:
> I'm happy to just see it removed.

Did this.

scpeters:
> Ideally we'd probably calculate the reverse dependencies for all the reverse dependencies we will install and test (rather than just the @formulae) but that may be annoyingly slow. Thoughts?

Yes, that would be slow. I'll submit a new PR if I think of a better approach.

sjackman:
Precomputing and storing (caching on disk) a graph (DAG) of dependencies could be useful in many situations, and speed up these sorts of operations. Thoughts?

MikeMcQuaid:
> Precomputing and storing (caching on disk) a graph (DAG) of dependencies could be useful in many situations, and speed up these sorts of operations.

@sjackman The main problem is that you'd need to regenerate the entire cache every time a formula changed and the slow bit is pulling all the formulae off disk. I'm open to being proved wrong here, though.

scpeters:
I was thinking to focus on the specific `brew install --only-dependencies` lines in [install_bottled_dependent](https://github.com/Homebrew/homebrew-test-bot/blob/ab1f932ae916ca1c41039c84093cf455d98a20cd/cmd/brew-test-bot.rb#L837-L838) when one of the already bottled `@formulae` might try to be installed (if we also restore the `brew uninstall` lines from this PR). We could store the names of all the already built bottles as a map from formula name to bottle name in `@formulae_already_bottled` and do something like the following pseudocode in `install_bottled_dependent` for `brew install --only-dependencies`:

~~~
deps = $(brew deps dependent)
deps.each do |d|
  if @formulae_already_bottled.has_key?(d)
    system("brew install #{@formulae_already_bottled[d]}")
  end
end
~~~

sjackman:
>> Precomputing and storing (caching on disk) a graph (DAG) of dependencies could be useful in many situations, and speed up these sorts of operations.

> @sjackman The main problem is that you'd need to regenerate the entire cache every time a formula changed and the slow bit is pulling all the formulae off disk. I'm open to being proved wrong here, though.

I was thinking of caching the graph in a JSON file that's either created by `brew update`, or alternatively created when the PR is merged, and stored in the GitHub repo of the tap.

MikeMcQuaid:
> I was thinking of caching the graph in a JSON file that's either created by `brew update`, or alternatively created when the PR is merged, and stored in the GitHub repo of the tap.

@sjackman It's going to be a slow operation (cannot be faster than `brew readall`) to regenerate that JSON file so it'll need to be opt-in. A better fit might be a flag that's used on `brew uses` or `brew deps` that generates the cache then. Check the `Library/Homebrew/description_cache_store.rb` and `Library/Homebrew/linkage_cache_store.rb` for some prior art on this.

MikeMcQuaid:
> We could store the names of all the already built bottles as a map from formula name to bottle name in `@formulae_already_bottled` and do something like the following pseudocode

Yeh, that could be a good call. We'll keep the bottle around so should be easy to install and uninstall that specifically.

PR #269, opened 2019-07-06
brew-test-bot: Add a `--skip-recursive-dependents` flag

issyl0:
- There are some _large_ formulae with many dependencies. One of these
  is `bzip2`. In those cases, we don't want to test all of its dependencies
  in one go because it times out on CI. We especially don't want to test
  all of its _recursive_ dependencies, as there are _lots_.
- We can perform manual tests of some formulae to ensure that we're
  keeping functionality, if the formuale that's depended on has made
  breaking changes to its libraries with its upgrade.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -88,6 +88,9 @@
 #:
 #:    If `--ci-upload` is passed, use the Homebrew CI bottle upload
 #:    options.
+#:
+#:    If `--skip-recursive-dependents` is passed, only test the direct
+#:    dependents.
 
 require "formula"
 require "formula_installer"
@@ -721,8 +724,10 @@ module Homebrew
       build_dependencies = dependencies - runtime_or_test_dependencies
       @unchanged_build_dependencies = build_dependencies - @formulae
 
+      uses_args = []
+      uses_args << "--recursive" unless ARGV.include?("--skip-recursive-dependents")
       dependents =
-        Utils.popen_read("brew", "uses", "--include-test", "--recursive", formula_name)
+        Utils.popen_read("brew", "uses", "--include-test", *uses_args, formula_name)
              .split("\n")
       dependents -= @formulae
       dependents = dependents.map { |d| Formulary.factory(d) }


cmd/brew-test-bot.rb
```diff
@@ -721,15 +723,17 @@ def setup_formulae_deps_instances(formula, formula_name)
       build_dependencies = dependencies - runtime_or_test_dependencies
       @unchanged_build_dependencies = build_dependencies - @formulae
 
-      dependents =
-        Utils.popen_read("brew", "uses", "--include-test", "--recursive", formula_name)
-             .split("\n")
-      dependents -= @formulae
-      dependents = dependents.map { |d| Formulary.factory(d) }
+      unless ARGV.include?("--skip-dependants")
```
issyl0: Helps if I’m consistent with spelling...

```suggestion
      unless ARGV.include?("--skip-dependents")
```

MikeMcQuaid:
> * There are some _large_ formulae with many dependencies. One of these
>   is `bzip2`. In those cases, we don't want to test all of its dependencies
>   in one go because it times out on CI. We especially don't want to test
>   all of its _recursive_ dependencies, as there are _lots_.

How about only testing the direct dependencies and skipping the recursive instead? My concern with this flag is that it will definitely result in breakages due to lack of testing.

issyl0:
@MikeMcQuaid Thanks, I've made a change that doesn't include `--recursive` if the skip flag is set, but still runs `brew uses --include-test formula`.

cmd/brew-test-bot.rb
```diff
@@ -721,8 +724,10 @@ def setup_formulae_deps_instances(formula, formula_name)
       build_dependencies = dependencies - runtime_or_test_dependencies
       @unchanged_build_dependencies = build_dependencies - @formulae
 
+      dependents_args = %w[--include-test]
```
sjackman:
```suggestion
      uses_args = []
```

cmd/brew-test-bot.rb
```diff
@@ -721,8 +724,10 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+      dependents_args << "--recursive" unless ARGV.include?("--skip-recursive-dependents")
```
sjackman:
```suggestion
      uses_args << "--recursive" unless ARGV.include?("--skip-recursive-dependents")
```

cmd/brew-test-bot.rb
```diff
@@ -721,8 +724,10 @@ def setup_formulae_deps_instances(formula, formula_name)
...
       dependents =
-        Utils.popen_read("brew", "uses", "--include-test", "--recursive", formula_name)
+        Utils.popen_read("brew", "uses", *dependents_args, formula_name)
```
sjackman:
```suggestion
        Utils.popen_read("brew", "uses", "--include-test", *uses_args, formula_name)
```

issyl0:
@sjackman Thanks, I've applied your suggestions.

cmd/brew-test-bot.rb
```diff
@@ -721,8 +724,10 @@ def setup_formulae_deps_instances(formula, formula_name)
       build_dependencies = dependencies - runtime_or_test_dependencies
       @unchanged_build_dependencies = build_dependencies - @formulae
 
+      uses_args = []
+      uses_args << "--recursive" unless ARGV.include?("--skip-recursive-dependents")
```
MikeMcQuaid: Could do something similarly here perhaps? https://github.com/Homebrew/homebrew-test-bot/blob/f7429c951c86edac45ab2e8bc85d410aae0deb1b/cmd/brew-test-bot.rb#L873

Not gonna block this on that suggestion (as it's not necessary but may speed things up).

MikeMcQuaid:
Thanks again @issyl0!

cmd/brew-test-bot.rb
sjackman: This line runs `brew uses --recursive formula_name` and then ignores the result. Do you know the purpose of this line?

sjackman: Can it be removed altogether?

scpeters: that line was last edited in https://github.com/Homebrew/homebrew-test-bot/pull/21

Since it's at the beginning of `def formula`, I'm guessing it lets you fail fast if there are any problems with formulae?

sjackman: I want back through the commit history for that line, and `brew uses formula_name` goes back to the dawn of time. The `--recursive` option was added, then removed for performance reasons, then added back in. I'm guessing it's to fail fast as well, but since we run `brew readall`, it seems a bit redundant.

scpeters: I suppose we could try removing it in another branch and see if the code coverage changes?

sjackman: @MikeMcQuaid Thoughts?

MikeMcQuaid: I'm fine with seeing it removed as it's used for testing and being able to query the output rather than `test-bot` doing anything with it. This would be a case where a `brew uses` cache would make sense.

sjackman: I've opened PR https://github.com/Homebrew/homebrew-test-bot/pull/270 to remove this call to `brew uses --recursive`.

PR #270, opened 2019-07-09
Remove the call to brew uses --recursive

sjackman:
The result of this call to brew uses --recursive is ignored.


MikeMcQuaid:
Seems fair 👍

MikeMcQuaid:
> The result of this call to brew uses --recursive is ignored.

Just to be explicit on this, though: it's not ignored as it'll be marked as failing if `uses` fails due to e.g. a bad formula that depends on it. This doesn't seem to be an issue, though, so should be safe to remove.

PR #271, opened 2019-07-11
ci-upload: redact token in the log

xu-cheng:
Depends on https://github.com/Homebrew/brew/pull/6291


xu-cheng:
This PR should also be merged. Noted due to the usage of `ENV.clear_sensitive_environment!`, we have to specify tokens here.

MikeMcQuaid:
Cool, thanks @xu-cheng!

sjackman:
Thank you, @xu-cheng !

iMichka:
Thanks @xu-cheng !

PR #279, opened 2019-09-25
Tap REQUIRED_TAPS when --tap option is specified

dawidd6:
This change is mostly needed on linux. Tapping `linuxbrew/xorg` is required for some formulas and it is not done when `brew test-bot` is run with `--tap` option specified.

I didn't really give this change a test run, but if I understand the code well, it should do the job and prevent situations like this one:

https://github.com/dawidd6/homebrew-tap/commit/d6f07e80d6178260e28870f93a24a68c8cf8e43a/checks

where whole build fails because of this:
```
==> brew linkage --test dawidd6/tap/neofetch
Error: No available formula with the name "linuxbrew/xorg/xorg" 
Please tap it and then try again: brew tap linuxbrew/xorg
==> FAILED
```

and manual tapping of `linuxbrew/xorg` before build seems to fix it:

https://github.com/dawidd6/homebrew-tap/commit/4d549e4e54c34da15a142de64506888dc1d77b9e/checks

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -596,6 +596,13 @@ module Homebrew
       @test_brew = (!@tap || @test_bot_tap) &&
                    (@formulae.empty? || @test_default_formula)
 
+      unless @test_brew
+        installed_taps = Tap.select(&:installed?).map(&:name)
+        (REQUIRED_TAPS - installed_taps).each do |tap|
+          test "brew", "tap", tap
+        end
+      end
+
       puts <<~EOS
 
         Formula changes to be tested:


cmd/brew-test-bot.rb
```diff
@@ -1090,6 +1090,11 @@ def homebrew
         test "brew", "man", "--fail-if-changed"
         test "brew", "style"
       elsif @tap
+        installed_taps = Tap.select(&:installed?).map(&:name)
```
MikeMcQuaid: I think this should be done only when `@test_brew` isn't set (rather than needing a `@tap`) and might make more sense in `def download`, thoughts?

dawidd6: It sure might be a better way of doing that. Just force-pushed this change.

iMichka:
In `linuxbrew-core` we tap `xorg` in the azure ci yaml file. And when the builds were run on circle-ci or travis back then, it was the same.
@sjackman was there any particular reason why it was done like this, instead of letting the test-bot auto-tap what's needed?

sjackman:
> In linuxbrew-core we tap xorg in the azure ci yaml file. And when the builds were run on circle-ci or travis back then, it was the same.
> @sjackman was there any particular reason why it was done like this, instead of letting the test-bot auto-tap what's needed?

I think just laziness. I like this fix.

sjackman:
Thank you for this fix, Dawid! 🏅

iMichka:
> I think just laziness. I like this fix.

No issue, just wanted to be sure there was no reason against the change. I like it too.

Thanks!

PR #280, opened 2019-09-25
Use color in Github Actions

dawidd6:
Looks like colorized text is showing up just fine on Github Actions:

https://github.com/dawidd6/homebrew-test/commit/585717b67f9eda62a79c69e699f50b21fff9b397/checks

I think it's worth it to make this change.


MikeMcQuaid:
Nice spot! this makes things much more readable.

Issue #282, opened 2019-09-26
Github Actions and ARGV


dawidd6:
I'm trying to get `test-bot` up and running for my own tap and it seems to work when `GITHUB_ACTIONS` environment variable is not set. When it is set, `test-bot` is automatically appending some options and it looks like they break the formula detection.

Logs from workflow run when `GITHUB_ACTIONS` is **not** set:
```
2019-09-26T00:27:10.6753530Z ==> Tapping homebrew/test-bot
2019-09-26T00:27:10.6955133Z Cloning into '/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-test-bot'...
2019-09-26T00:27:11.3288993Z Tapped 1 command and 1 formula (50 files, 260.6KB).
2019-09-26T00:27:20.0412462Z Homebrew/homebrew-test-bot fb05b68 (Merge pull request #279 from dawidd6/tap)
2019-09-26T00:27:20.0414495Z ARGV: --verbose --root-url=https://dl.bintray.com/dawidd6/bottles-tap
2019-09-26T00:27:20.0420232Z brew tap dawidd6/tap --full
2019-09-26T00:27:20.8428157Z ==> Tapping dawidd6/tap
2019-09-26T00:27:20.8632130Z Cloning into '/home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/dawidd6/homebrew-tap'...
2019-09-26T00:27:21.2177880Z Tapped 7 formulae (34 files, 214.6KB).
2019-09-26T00:27:21.6552743Z 
2019-09-26T00:27:21.6554703Z ==> brew pull --clean https://github.com/dawidd6/homebrew-tap/pull/31/checks
2019-09-26T00:27:22.7760544Z ==> Fetching patch 
2019-09-26T00:27:22.7761538Z Patch: https://github.com/dawidd6/homebrew-tap/pull/31.patch
2019-09-26T00:27:24.8099061Z ==> Applying patch
2019-09-26T00:27:24.8364312Z Applying: neofetch: test
2019-09-26T00:27:24.9047579Z Warning: dawidd6/tap/neofetch has a bottle: do you need to update it with --bottle?
2019-09-26T00:27:24.9252666Z ==> Patch changed:
2019-09-26T00:27:24.9466862Z  Formula/neofetch.rb | 1 +
2019-09-26T00:27:24.9467143Z  1 file changed, 1 insertion(+)
[...]

PR #284, opened 2019-09-26
test-bot: better handle missing CI variables.

MikeMcQuaid:
Print a big warning so people understand what's going on.

CC @dawidd6 for thoughts.

Closes: #282


MikeMcQuaid:
https://github.com/Homebrew/homebrew-test-bot/pull/284/files#diff-8214fa507ee396ae17e4b9c27cb881a2R463-R467 confirmed this message is not printed on any of the CI jobs so: merging.

PR #286, opened 2019-09-27
Allow non-core taps to create bintray packages

dawidd6:
Normal bintray users need to specify `licenses` and `vcs_url` fields when creating packages.

I think `HOMEBREW_BOTTLE_DOMAIN` can be replaced by the url I provided, cause it works for everyone, "homebrew", "linuxbrew" and others.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -1525,13 +1525,14 @@ module Homebrew
       bintray_repo = bottle_hash["bintray"]["repository"]
       bintray_packages_url =
         "https://api.bintray.com/packages/#{bintray_org}/#{bintray_repo}"
+      bintray_bottle_url = "https://dl.bintray.com/#{bintray_org}"
 
       rebuild = bottle_hash["bottle"]["rebuild"]
 
       bottle_hash["bottle"]["tags"].each do |tag, _tag_hash|
         filename = Bottle::Filename.new(formula_name, version, tag, rebuild)
         bintray_url =
-          "#{HOMEBREW_BOTTLE_DOMAIN}/#{bintray_repo}/#{filename.bintray}"
+          "#{bintray_bottle_url}/#{bintray_repo}/#{filename.bintray}"
         filename_already_published = if ARGV.include?("--dry-run")
           puts "curl -I --output /dev/null #{bintray_url}"
           false
@@ -1561,6 +1562,8 @@ module Homebrew
           unless package_exists
             package_blob = <<~EOS
               {"name": "#{bintray_package}",
+               "vcs_url": "#{bintray_bottle_url}",
+               "licenses": ["MIT", "GPL-2.0", "GPL-3.0", "BSD"],
                "public_download_numbers": true,
                "public_stats": true}
             EOS


cmd/brew-test-bot.rb
```diff
@@ -1561,6 +1562,8 @@ def test_ci_upload(tap)
           unless package_exists
             package_blob = <<~EOS
               {"name": "#{bintray_package}",
+               "vcs_url": "#{bintray_bottle_url}",
+               "licenses": ["MIT", "GPL-2.0", "GPL-3.0", "BSD"],
```
MikeMcQuaid: These two lines aren't correct so I'm a bit dubious about having them hardcoded.

cmd/brew-test-bot.rb
```diff
@@ -1525,13 +1525,14 @@ def test_ci_upload(tap)
       bintray_repo = bottle_hash["bintray"]["repository"]
       bintray_packages_url =
         "https://api.bintray.com/packages/#{bintray_org}/#{bintray_repo}"
+      bintray_bottle_url = "https://dl.bintray.com/#{bintray_org}"
```
MikeMcQuaid: What about `"https://#{bintray_org}.bintray.com"`?

dawidd6: It does not work for everyone.
https://dawidd6.bintray.com/bottles-tap/keybase-cli-4.4.1.x86_64_linux.bottle.tar.gz
vs
https://bintray.com/dawidd6/bottles-tap/keybase-cli-4.4.1.x86_64_linux.bottle.tar.gz

cmd/brew-test-bot.rb
dawidd6: Yea, I know, but these fields are required to create a package on bintray as normal user, so what can we fill them with?

MikeMcQuaid: I'm not sure but if they are required fields I don't think Bintray will look kindly at us setting them by default to incorrect values.

cmd/brew-test-bot.rb
MikeMcQuaid: I don't want to change the default URL that does work for Homebrew.

jonchang:
This has bit me before... I think better messaging is a partial solution here with respect to the license field. Or, if we ever do merge the licenses project that would be an out as well. Agree with Mike that we shouldn't be setting these for the user.

cmd/brew-test-bot.rb
sjackman: I suggest…
```
               "vcs_url": "https://github.com/Homebrew/brew",
               "licenses": ["BSD-2"],
```

cmd/brew-test-bot.rb
sjackman: @MikeMcQuaid Why do you prefer `ORG.bintray.com` over `dl.bintray.com/ORG`? The former only works for Homebrew because we're sponsored. I prefer the latter that works for everyone.

MikeMcQuaid: It reads nicer to me but mostly:

> I don't want to change the default URL that does work for Homebrew.

sjackman: My understanding is that Homebrew would like to improve the experience for third party taps and bottling on third party taps. This change would is a step toward that goal. One option would be to keep `homebrew.bintray.com` for official taps and use `dl.bintray.com/homebrew` for third party taps.

dawidd6: there is also `linuxbrew.bintray.com`

dawidd6:
Closing, cause Bintray said we should not set static values as license and vcs url.

sjackman:
Supporting storing bottles as GitHub release artifacts for third-party taps could be a viable workaround, if someone sufficiently motivated wanted to do the work.

PR #287, opened 2019-09-27
Check if merge commit and CoreTap and OS.linux

dawidd6:
Fixes problems with detection of modified formulae on taps other than core.


MikeMcQuaid:
Makes sense, thanks @dawidd6! CC @sjackman @iMichka FYI.

sjackman:
Sounds good. Thanks for the heads up.

PR #291, opened 2019-10-20
test-bot: remove duplicated --dry-run

dawidd6:


issyl0:
Ooh this is fun:

```
expected block to not output to stderr, but output "/usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.3/lib/ruby/2.6.0/rubygems.rb:661: [BUG] Segmentation fault at 0x0000000000000018\nruby 2.6.3p62 (2019-04-16 revision 67580)
```

dawidd6:
Is it because of that change?

issyl0:
If docs changes are causing segfaults then I'd be very worried!

issyl0:
Looks to be the same as https://github.com/Homebrew/brew/issues/6579#issuecomment-542088400.

PR #292, opened 2019-10-20
git rm .travis.yml

issyl0:
- Each PR to this repo runs five status checks. Four of those checks run
  `brew test-bot` on various CI infrastructure.
- We have: Travis CI (Linux), Azure Pipelines (Linux), GitHub Actions
  (macOS and Linux). This feels excessive.
- Travis CI was (re)introduced in
  73b160c16fd0e573fb91d8e0f29d5267e80cc462 having been removed in
  48e567b25ff269721f86e2503b9a5370ccf7c97a, and it's not speedy. That
  commit's description says "linux-only configuration to ensure we don't
  break things", which has been done by Azure and GitHub Actions since
  then.
- I propose removing the Travis checks, as we have two other CI
  providers doing Linux checks for us.

---

Am I missing any context as to why this shouldn't be done?


MikeMcQuaid:
> Am I missing any context as to why this shouldn't be done?

Yes, sorry: we still have people using `brew test-bot` on all these platforms and we have custom code for all these platforms so we want to ensure we don't break it.

PR #295, opened 2019-11-06
test-bot: check build dependents.

MikeMcQuaid:
This should ensure that we don't e.g. update `go` or `ghc` where most dependents are `:build` and break those formulae building from source.

As requested by @fxcoudert.

### DIFF
--- a/cmd/brew-test-bot.rb
+++ b/cmd/brew-test-bot.rb
@@ -755,7 +755,7 @@ module Homebrew
       uses_args = []
       uses_args << "--recursive" unless ARGV.include?("--skip-recursive-dependents")
       dependents =
-        Utils.popen_read("brew", "uses", "--include-test", *uses_args, formula_name)
+        Utils.popen_read("brew", "uses", "--include-build", "--include-test", *uses_args, formula_name)
              .split("\n")
       dependents -= @formulae
       dependents = dependents.map { |d| Formulary.factory(d) }


fxcoudert:
This will make updates of `go`, `rust`, `ghc` a lot harder. But it's a good thing, because we will have less breakage down the line.

MikeMcQuaid:
@fxcoudert Thinking more about this PR I fear it may have been mostly pointless. 

> This should ensure that we don't e.g. update `go` or `ghc` where most dependents are `:build` and break those formulae building from source.

Because we only test the **bottles of dependents** this will only have tested the cases where there was no bottle or a `revision` was present (which would have been built from source regardless) so I'm pretty sure this added a bunch of CI time for literally no reason 😭.

What _could_ work is generating these dependents and building them from source. I'd suggest this will make the build times dramatically increase for anything with a non-trivial number of build dependencies so it may be instead worth taking a random sample of build dependencies and building them (to identify cases where e.g. our `go`/`rust` has badly broken other projects).

scpeters:
> it may be instead worth taking a random sample of build dependencies and building them (to identify cases where e.g. our go/rust has badly broken other projects)

A random sample is an interesting idea.

Bo98:
Revisiting the point of:

> What could work is generating these dependents and building them from source. I'd suggest this will make the build times dramatically increase for anything with a non-trivial number of build dependencies so it may be instead worth taking a random sample of build dependencies and building them

Ultimately, I feel that it depends a little on the formula.

For example, testing the dependents of `cmake` is not particularly useful. Due to the amount of effort upstream put into backwards compatibility, it is extremely rare for a CMake update to have broken a build. Particularly when you combine that with the number of formulae depending on `cmake`, there's little benefit to rebuilding dependents of `cmake`.

On the other side, we should really be rebuilding all dependents on updates like `ocaml`. OCaml updates are extremely likely to break old builds, typically because of version pinning. Every formula I have touched in the last few months that has depended on OCaml has been broken.

There are similar issues with `ghc` (and `cabal-install`) as demonstrated here: https://github.com/Homebrew/homebrew-core/issues/48604

MikeMcQuaid:
> On the other side, we should really be rebuilding all dependents on updates like `ocaml`. OCaml updates are extremely likely to break builds, typically because of version pinning. Every formula I have touched in the last few months that has depended on OCaml has been broken.
> 
> There are similar issues with `ghc` (and `cabal-install`, though that was because of a major update) as demonstrated here: [Homebrew/homebrew-core#48604](https://github.com/Homebrew/homebrew-core/issues/48604)

I'm up for, in that case, somehow having a way of whitelisting the stuff we want to do this on. It could be a formula DSL or just (while we get started, at least) a whitelist in the `test-bot` code itself. 

Thoughts?

Bo98:
It'll have to be something like that, yeah.

There is still a question around build time increasing. OCaml's 28 dependents isn't an insignificant number considering compile times for OCaml projects aren't that great, but if it is anything like 4.07.0, 4.08.0 and 4.09.0 were, most of those 28 probably require tweaks anyway.

OCaml 4.10.0 was released 3 days ago, so that's a possible candidate for trialling things.

If rebuild time is a significant issue, we would need to ask ourselves what would be the best approach to split up the work into managable chunks for problematic formulae like OCaml, like for example using the Python 3.8 approach.

MikeMcQuaid:
Have opened https://github.com/Homebrew/homebrew-test-bot/pull/329, let me know what you think.

PR #299, opened 2019-11-20
Move Test#homebrew logic to GitHub Actions.

MikeMcQuaid:
This will improve the output and ease of parsing while allowing code to be removed from brew-test-bot which relates only to a single repo.

Homebrew/brew companion PR: https://github.com/Homebrew/brew/pull/6760.


iMichka:
Nice cleanup. I always found it weird to have this code in brew-test-bot.

PR #306, opened 2019-12-20
lib/test: cleanup during build.

MikeMcQuaid:
Ensure that every time we install a new formula or bottled dependent that we check the state of the disk usage and nuke HOMEBREW_CACHE if it has got too full.

### DIFF
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -570,6 +570,8 @@ module Homebrew
     end
 
     def install_bottled_dependent(dependent)
+      cleanup_during
+
       unless dependent.installed?
         test "brew", "fetch", "--retry", dependent.name
 
@@ -606,6 +608,8 @@ module Homebrew
     end
 
     def formula(formula_name)
+      cleanup_during
+
       @category = "#{__method__}.#{formula_name}"
 
       formula = Formulary.factory(formula_name)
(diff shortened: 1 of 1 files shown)


Bo98:
Interesting solution. I think this should work - I guess we'll soon find out!

I'm always cautious parsing command line output given there's likely edge cases (e.g. `/home`) but I think it will be ok for this. Shame that macOS doesn't have `--output=pcent`.

Does anything appear in the console log when this cleanup is triggered? Doesn't really matter - just interested.

MikeMcQuaid:
> Interesting solution. I think this should work - I guess we'll soon find out!

Hah yup 😅 

> I'm always cautious parsing command line output given there's likely edge cases (e.g. `/home`) but I think it will be ok for this. Shame that macOS doesn't have `--output=pcent`.

Yeh, I also looked for a better solution.

> Does anything appear in the console log when this cleanup is triggered? Doesn't really matter - just interested.

It wouldn't so I've added a `test "brew", "cleanup", "--prune=0"` before the `HOMEBREW_CACHE.children.each(&:rmtree)`. That'll mean the `==> brew cleanup --prune=0` will show up in the logs much in the same way `==> brew cleanup --prune=3` shows up currently at the end.

Let me know after this is merged if it seems sufficient or not. We could also consider clearing e.g. installed formulae and/or `/tmp` if necessary.

Bo98:
> Let me know after this is merged if it seems sufficient or not. We could also consider clearing e.g. installed formulae and/or `/tmp` if necessary.

Will do. :+1:

iMichka:
Let's try this :)

PR #311, opened 2020-01-07
lib/test: prioritise formula names over commit hashes.

MikeMcQuaid:
If needed a longer hash can be used. Fixes detection of `abcde`.


fxcoudert:
Thanks!

PR #312, opened 2020-01-09
Remove support for Travis and Circle CI.

MikeMcQuaid:
These are no longer used by any official Homebrew taps so nudge people towards GitHub Actions or Azure Pipelines instead.


maxim-belkin:
May I suggest to factor out all CIs into their own files which are sourced when / if necessary? The fact that Homebrew no longer uses CircleCI / Travis is fine but it feels strange that one has to manually look for pieces that correspond to a particular CI.

MikeMcQuaid:
> May I suggest to factor out all CIs into their own files which are sourced when / if necessary?

I'm open to further refactoring of `brew test-bot` but don't plan on doing so in the near future myself. Also, given Jenkins support will 💀 hopefully in 2020 and GitHub Actions and Azure Pipelines perform nearly identically the divergence should be minimal.

sjackman:
RIP

PR #318, opened 2020-01-21
Add option brew test-bot --ci-upload --publish

sjackman:
If `--publish` is passed, publish the uploaded bottles.


MikeMcQuaid:
Makes sense to me, thanks @sjackman!

PR #326, opened 2020-02-08
Add --override option to brew test-bot --ci-upload

dawidd6:
With this option passed, `test-bot` will force overwrite already
uploaded and published bottles on Bintray.

Needed to make a separate API call for publishing,
because specifying `?override` with `?publish` in the URL didn't work as intended.

Tested cases:
- `--publish && --override` (works, overwrites bottles and publishes them)
- `--publish` (works as before)
- `--override` (works, overwrites bottles without publishing)

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -63,6 +63,8 @@ module Homebrew
              description: "automatically pick one of the Homebrew CI options based on the environment. Implies `--cleanup`: use with care!"
       switch "--ci-upload",
              description: "use the Homebrew CI bottle upload options."
+      switch "--override",
+             description: "override already uploaded and published bottles."
       switch "--publish",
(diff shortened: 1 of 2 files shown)


sjackman:
What maintainer workflow that causes a bottle to published on Bintray but its bottle commit not pushed to `master`? This case should I hope be relatively rare. In that rare case, I feel that in that rare case the maintainer can manually delete the bottles using the Bintray web interface.

dawidd6:
Okay then. Was just testing a workflow, where CI pushes the bottle commit back to the PR and overwriting could be useful in that case.

sjackman:
You could possibly make the workflow work with unpublished bottles. Unpublished bottles may be overwritten by default with the `override` option, which is only meaningful for published bottles.

PR #327, opened 2020-02-20
Add `HOMEBREW_CURL_RETRIES` to make downloads/audit more robust.

reitermarkus:
Retry `curl` commands three times to reduce the chance of intermittent network issues causing CI to fail.


MikeMcQuaid:
I'm game for this but maybe it should be a Homebrew/brew default? Thoughts?

PR #328, opened 2020-02-24
Add logic to handle self-hosted runners

jonchang:
* self hosted runners need all of `cleanup_after` since the VMs are not ephemeral
* self hosted runners should modify $HOME to the runner working directory to avoid polluting the VM home directory

See https://github.com/Homebrew/homebrew-core/pull/50620 for how this affects self-hosted runners on actions.

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -83,6 +83,7 @@ module Homebrew
 
   def setup_argv_and_env
     jenkins = !ENV["JENKINS_HOME"].nil?
+    github_actions_self_hosted = !ENV["GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED"].nil?
 
     github_actions = !ENV["GITHUB_ACTIONS"].nil?
     if github_actions
@@ -107,7 +108,8 @@ module Homebrew
        ARGV.include?("--ci-testing")
       ARGV << "--cleanup"
       ARGV << "--test-default-formula"
-      ARGV << "--local" << "--junit" if jenkins
+      ARGV << "--local" if github_actions_self_hosted || jenkins
+      ARGV << "--junit" if jenkins
     end
 
     ARGV << "--verbose" if ARGV.include?("--ci-upload")
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -882,7 +882,7 @@ module Homebrew
       @category = __method__
       return if @skip_cleanup_after
 
-      if ENV["HOMEBREW_AZURE_PIPELINES"] || ENV["HOMEBREW_GITHUB_ACTIONS"]
+      if ENV["HOMEBREW_GITHUB_ACTIONS"] && !ENV["GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED"]
         # don't need to do post-build cleanup unless testing test-bot itself.
         return if @tap.to_s != "homebrew/test-bot"
       end


cmd/test-bot.rb
```diff
@@ -83,10 +83,12 @@ def test_bot
 
   def setup_argv_and_env
     jenkins = !ENV["JENKINS_HOME"].nil?
+    self_hosted = !ENV["HOMEBREW_SELF_HOSTED_RUNNER"].nil?
```
MikeMcQuaid:
```suggestion
    github_actions_self_hosted = !ENV["GITHUB_ACTIONS_SELF_HOSTED_RUNNER"].nil?
```

cmd/test-bot.rb
```diff
@@ -83,10 +83,12 @@ def test_bot
...
 
     github_actions = !ENV["GITHUB_ACTIONS"].nil?
     if github_actions
-      ARGV << "--verbose" << "--ci-auto" << "--no-pull"
+      ARGV << "--verbose" << "--ci-auto"
+      ARGV << "--no-pull" unless self_hosted
```
MikeMcQuaid:
```suggestion
      ARGV << "--no-pull" unless github_actions_self_hosted
```

but I think it'd be preferable to avoid needing `--no-pull` here if possible.

cmd/test-bot.rb
```diff
@@ -107,7 +109,8 @@ def setup_argv_and_env
        ARGV.include?("--ci-testing")
       ARGV << "--cleanup"
       ARGV << "--test-default-formula"
-      ARGV << "--local" << "--junit" if jenkins
+      ARGV << "--local" if self_hosted || jenkins
```
MikeMcQuaid:
```suggestion
      ARGV << "--local" if github_actions_self_hosted || jenkins
```

lib/test.rb
```diff
@@ -882,7 +882,7 @@ def cleanup_after
       @category = __method__
       return if @skip_cleanup_after
 
-      if ENV["HOMEBREW_AZURE_PIPELINES"] || ENV["HOMEBREW_GITHUB_ACTIONS"]
+      if ENV["HOMEBREW_GITHUB_ACTIONS"] && !ENV["HOMEBREW_SELF_HOSTED_RUNNER"]
```
```suggestion
      if ENV["HOMEBREW_GITHUB_ACTIONS"] && !ENV["GITHUB_ACTIONS_SELF_HOSTED_RUNNER"]
```

jonchang: This is a runner that Homebrew administers, not Github, so I don't think it makes sense to call it a "github" variable.

cmd/test-bot.rb
jonchang: Either:

* `brew test-bot` grabs the correct set of commits with `brew pull`, or
* we duplicate that logic in the workflow.

Re: https://github.com/Homebrew/homebrew-core/pull/50468#discussion_r383032833

This cannot use `actions/checkout` because, like the Jenkins setup, the VMs are persistent. Using `actions/checkout` is only possible on Linux [because we remove `$(brew --repo homebrew/core)`](https://github.com/Homebrew/linuxbrew-core/blob/f20b3f97559466a46cb87cb392201ea90c509b37/.github/workflows/build-bottles.yml#L15-L20) and the runners are ephemeral so it doesn't matter if the tap folder gets taken out.

cmd/test-bot.rb
jonchang: See https://github.com/Homebrew/homebrew-test-bot/pull/328#discussion_r383227610

cmd/test-bot.rb
MikeMcQuaid: > This cannot use `actions/checkout` because, like the Jenkins setup, the VMs are persistent.

Can you elaborate a bit on this? Would love to get e.g. `brew test-bot` to a point where we can use actions/checkout.

jonchang: I thought I already did, in the next sentence. What do you need clarification about?

sjackman: You could `rm -rf $(brew --repo homebrew/core)` and then run `actions/checkout`.

jonchang: A full clone is something like 2-4 minutes per CI job. I didn’t think that was an acceptable trade off given we avoid it in the dispatch build bottle job on Linux, and that we changed the Dockerfile to include full clones of homebrew/core.

sjackman: Good point. I agree!
In that case I'd suggest
```
cd $(brew --repo homebrew/core) && brew reset --hard $GITHUB_SHA`
```
either in the GitHub Action before running `brew test-bot` or in `brew test-bot`.

MikeMcQuaid: @jonchang I don't think I understand why persistence means we can't use it. Is it because it can only do a full clone every time and not/never e.g. update an existing clone?

lib/test.rb
MikeMcQuaid: All `HOMEBREW_` variables show up in `brew config` unless specifically blacklisted. It's not configuration for Homebrew itself, really, and we don't need the `HOMEBREW_` prefix because `test-bot` doesn't have environment filtering. Could do `GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED_RUNNER` or something?

cmd/test-bot.rb
jonchang: Shaun: I tried that already. You need to do a `git fetch origin refs/pull/pr/merge` as the existing repository doesn’t know what object `github.sha` is because there’s no ref that points to it. 

Mike: I didn’t see anything in the documentation that said actions/checkout could do that. It only refers to new clones (shallow or full).

MikeMcQuaid: Thanks @jonchang!

sjackman:
```sh
cd $(brew --repo homebrew/core)
git fetch origin "$GITHUB_REF"
git checkout master
git reset --hard "$GITHUB_SHA"
```
?

jonchang: Yeah, this works: https://github.com/Homebrew/linuxbrew-core/pull/19715

`actions/checkout@v2` to an existing directory fails: https://github.com/Homebrew/linuxbrew-core/runs/468679652

MikeMcQuaid: Thanks for investigating ❤️

PR #331, opened 2020-02-27
Use /usr/bin/git

### LINKED ISSUE #330: Test bot fails when brew git is tested
We've been merging failing `git` builds for far too long - ideally we wouldn't be merging anything that's red. [Here is an example failure from git](https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/57713/version=mojave/console) - please ignore the `brew test` failure here as that is something specific to that version bump. I'm talking about this at the end:

```
==> brew uninstall --force gettext pcre2

==> git -C homebrew/homebrew-core checkout master -f
==> FAILED
dyld: Library not loaded: /usr/local/opt/pcre2/lib/libpcre2-8.0.dylib
  Referenced from: /usr/local/bin/git
  Reason: image not found
dyld: Library not loaded: /usr/local/opt/pcre2/lib/libpcre2-8.0.dylib
  Referenced from: /usr/local/bin/git
  Reason: image not found

==> git -C homebrew/homebrew-core reset --hard origin/master
==> FAILED
dyld: Library not loaded: /usr/local/opt/pcre2/lib/libpcre2-8.0.dylib
  Referenced from: /usr/local/bin/git
  Reason: image not found
```

homebrew-test-bot uses whatever git is in the path (or Xcode.app if there's nothing). At the point of failure above, brew git is installed but non-functional because `pcre2` is uninstalled. This uninstall is intentional as if there is subsequent formulae being built then we want a cleaner environment, and to also make better use of disk space. The `git` formula isn't uninstalled to avoid rebuilds if there was a dependent due to be tested in the same job.

In addition to the situation above, the situation is worsened in the event that a git build is cancelled. This doesn't have to be manual - adding another commit to the pull request or force-pushing will trigger a cancellation. If it is cancelled in the middle of cleanup, the brew git install might not be wiped. This issue resulted in `mojave_orka_1` having to be taken offline earlier today as homebrew-test-bot currently cannot reach the point of wiping previous installs without a functional git. It requires manually going in and cleaning it up yourself - something I had hoped to have done by now if I wasn't having issues connecting to the Orka service endpoint.

Here is me brainstorming possible solutions:

* Always directly invoke system git. There's a few possible approaches:
  - Directly use `/usr/bin/git` in homebrew-test-bot.
    * This will not work on systems without the CLT. That's not an issue for our machines but may be for others using this tap.
    * It could somewhere else on Linux like `/usr/local/bin/git`.
  - Have homebrew-test-bot do the above and also search other locations (including Xcode.app) in the path that are not a part of Homebrew:
    * This would just end up as a reimplementation of the git shim so I'm not a fan of this approach of duplicating code.
  - Modify the git shim in brew to have an option or environment variable to alter the search and make brew git a last resort.
    * I suggest not just making this the default to reduce the chance of breakages.
      - Notably, `git-filter-repo` requires a newer git so its tests may fail if we change the default as it may go through the shim.
      - On Linux, the system git situation seems to be hit or miss at times so best to just avoid that.
    * How rigourous would we want this? A simple check for a `git` binary in `HOMEBREW_PREFIX/"bin"` or more advanced checks to see whether this is actually a brew-installed git rather than one installed from another location?
      - The shim is a shell script, so doing a more advanced check is a little more tricky than in Ruby and probably requires symlink checking.
    * Adding an option or environment variable feels a bit like a hack specific to homebrew-test-bot, but maybe there's some legitimate use cases in other scenarios.
    * Nevertheless, it's currently my preferred option on this list I reckon.
* Avoid invoking `git` in `cleanup_after` before the brew git is wiped.
  - This approach might not fix all cases. The wipe only occurs at the very end. What if another formula is tested after git in the same job but said formula invokes `git` as a part of its download & install process?
* git-specific hacks in homebrew-test-bot to uninstall brew git immediately
  - Problematic if a dependent is built in the same job - need to rebuild again.
  - Any hack like this probably won't even cover all edge cases - what if that dependent then uses git to download its sources (which should be after git is reinstalled again)?
* Other options I haven't thought of?

I can do some work on this and send a pull request if you want.

Bo98:
Fixes #330.

Disclaimer: test.rb is pretty much entirely untested. Testing was limited to `brew test-bot --ci-upload --dry-run` and `brew style homebrew/test-bot`.

### DIFF
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -909,13 +909,15 @@ module Homebrew
       return if @skip_cleanup_before
       return unless Homebrew.args.cleanup?
 
-      clear_stash_if_needed(@repository)
-      quiet_system @git, "-C", @repository, "am", "--abort"
-      quiet_system @git, "-C", @repository, "rebase", "--abort"
+      unless @test_bot_tap
+        clear_stash_if_needed(@repository)
+        quiet_system @git, "-C", @repository, "am", "--abort"
+        quiet_system @git, "-C", @repository, "rebase", "--abort"
 
-      unless Homebrew.args.no_pull?
-        checkout_branch_if_needed(@repository)
-        reset_if_needed(@repository)
+        unless Homebrew.args.no_pull?
+          checkout_branch_if_needed(@repository)
+          reset_if_needed(@repository)
+        end
       end
 
       Pathname.glob("*.bottle*.*").each(&:unlink)
@@ -948,8 +950,10 @@ module Homebrew
       end
 
       if Homebrew.args.cleanup?
-        clear_stash_if_needed(@repository)
-        reset_if_needed(@repository)
+        unless @test_bot_tap
+          clear_stash_if_needed(@repository)
+          reset_if_needed(@repository)
+        end
 
         test "brew", "cleanup", "--prune=3"
 
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -21,7 +21,7 @@ module Homebrew
 
     attr_reader :log_root, :category, :name, :steps
 
-    def initialize(argument, tap: nil, skip_setup: false, skip_cleanup_before: false, skip_cleanup_after: false)
+    def initialize(argument, tap:, git:, skip_setup: false, skip_cleanup_before: false, skip_cleanup_after: false)
       @hash = nil
       @url = nil
       @formulae = []
@@ -36,6 +36,7 @@ module Homebrew
       else
         CoreTap.instance.path
       end
+      @git = git
       @skip_setup = skip_setup
       @skip_cleanup_before = skip_cleanup_before
       @skip_cleanup_after = skip_cleanup_after
@@ -47,8 +48,8 @@ module Homebrew
         @pr_url = @url if pr
       elsif canonical_formula_name = safe_formula_canonical_name(argument)
         @formulae = [canonical_formula_name]
-      elsif quiet_system("git", "-C", @repository, "rev-parse",
-                             "--verify", "-q", argument)
+      elsif quiet_system(@git, "-C", @repository, "rev-parse",
+                               "--verify", "-q", argument)
         @hash = argument
       else
         raise ArgumentError,
@@ -76,17 +77,17 @@ module Homebrew
     end
 
     def current_sha1
-      Utils.popen_read("git", "-C", @repository,
-                              "rev-parse", "--short", "HEAD").strip
+      Utils.popen_read(@git, "-C", @repository,
+                             "rev-parse", "--short", "HEAD").strip
     end
 
     def diff_formulae(start_revision, end_revision, path, filter)
       return unless @tap
 
       Utils.popen_read(
-        "git", "-C", @repository,
-               "diff-tree", "-r", "--name-only", "--diff-filter=#{filter}",
-               start_revision, end_revision, "--", path
+        @git, "-C", @repository,
+              "diff-tree", "-r", "--name-only", "--diff-filter=#{filter}",
+              start_revision, end_revision, "--", path
       ).lines.map do |line|
         file = Pathname.new line.chomp
         next unless @tap.formula_file?(file)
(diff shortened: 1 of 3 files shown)


Bo98:
Huh, that `brew test-bot --ci-upload --dry-run` output certainly doesn't match mine.

lib/test_bot.rb
```diff
@@ -21,6 +21,8 @@ module Homebrew
   module TestBot
     module_function
 
+    GIT = "/usr/bin/git".freeze
```
MikeMcQuaid:
```suggestion
    GIT = "/usr/bin/git"
```

already frozen by `# frozen_string_literal: true` above.

lib/test_bot.rb
```diff
@@ -146,17 +148,17 @@ def test_ci_upload(tap)
...
         if Homebrew.args.dry_run?
           puts <<~EOS
-            git am --abort
-            git rebase --abort
-            git checkout -f master
-            git reset --hard origin/master
+            #{GIT} am --abort
+            #{GIT} rebase --abort
+            #{GIT} checkout -f master
+            #{GIT} reset --hard origin/master
```
MikeMcQuaid: I think these can probably stay as-is but up to you.

lib/test_bot.rb
```diff
@@ -283,12 +285,12 @@ def test_ci_upload(tap)
       return unless git_tag
 
       if Homebrew.args.dry_run?
-        puts "git tag --force #{git_tag}"
-        puts "git push --force #{remote} origin/master:master refs/tags/#{git_tag}"
+        puts "#{GIT} tag --force #{git_tag}"
+        puts "#{GIT} push --force #{remote} origin/master:master refs/tags/#{git_tag}"
```
MikeMcQuaid: I think these can probably stay as-is but up to you.

lib/test.rb
```diff
@@ -21,7 +21,7 @@ class Test
 
     attr_reader :log_root, :category, :name, :steps
 
-    def initialize(argument, tap: nil, skip_setup: false, skip_cleanup_before: false, skip_cleanup_after: false)
+    def initialize(argument, tap: nil, skip_setup: false, skip_cleanup_before: false, skip_cleanup_after: false, git: "git")
```
```suggestion
    def initialize(argument, git:, tap: nil, skip_setup: false, skip_cleanup_before: false, skip_cleanup_after: false)
```

Can also remove any other arguments that are always set.

lib/test_bot.rb
Bo98: I'm fine either way.

For some reason though it doesn't show as `/usr/bin/git` in GitHub actions and I don't quite get why.

MikeMcQuaid: Well spotted. Could it be it's not testing the right revision?

Bo98: It's weird since the previous step used it fine. Lets see if it happens again I guess.

Bo98: Or perhaps it's pulling it from master at some point?

Bo98: I see the issue. I'll try fix it.

lib/test.rb
```diff
@@ -908,10 +910,10 @@ def cleanup_before
...
-      quiet_system "git", "-C", @repository, "am", "--abort"
-      quiet_system "git", "-C", @repository, "rebase", "--abort"
+      quiet_system @git, "-C", @repository, "am", "--abort"
+      quiet_system @git, "-C", @repository, "rebase", "--abort"
 
-      unless Homebrew.args.no_pull?
+      unless Homebrew.args.no_pull? || @test_bot_tap
         checkout_branch_if_needed(@repository)
         reset_if_needed(@repository)
       end
```
MikeMcQuaid: Suggestion: do `unless @test_bot_tap` for all these lines.

lib/test.rb
```diff
@@ -947,7 +949,7 @@ def cleanup_after
 
       if Homebrew.args.cleanup?
         clear_stash_if_needed(@repository)
-        reset_if_needed(@repository)
+        reset_if_needed(@repository) unless @test_bot_tap
```
MikeMcQuaid: Suggestion: do `unless @test_bot_tap` for all these lines.

Bo98:
As someone who has never used RSpec before, hopefully I've done it correctly.

MikeMcQuaid:
Looks good, thanks @Bo98!

PR #333, opened 2020-03-04
test: skip dependents where requirements aren't satisfied

Bo98:
As per dicussion in #329.


MikeMcQuaid:
Good call, thanks again @Bo98!

PR #336, opened 2020-03-04
test-bot: use system curl everywhere

iMichka:
When building formulae, brewed curl may be in the path.
To be sure we are always using system curl, which is known to work,
we hardcode that path everywhere, as we did for git.


MikeMcQuaid:
Thanks @iMichka!

PR #340, opened 2020-03-15
test: avoid building formulae twice

Bo98:
Currently, it is possible that we build things twice. For example, if `ocaml` and `ocaml-findlib` are both modified in the same pull request, the sequence may be somehing like:

* Build `ocaml`.
* Test `ocaml` dependents, some of which will also depend on `ocaml-findlib`.
  - `ocaml-findlib` will be built by the `brew install --only-dependencies` step.
* `ocaml-findlib` is uninstalled and built once again, except with `--build-bottle` this time.
* Test `ocaml-findlib` dependents (which is just a subset of the `ocaml` dependents)
  - In this case there are no runtime dependencies, but there will be some duplication here for other formulae.

I initially saw this as just a build time inefficiency, but it turns out that this process has a serious effect on the bottles we produce: Homebrew/homebrew-core#51698. The `etc` and `var` part of our bottle generation process relies on there being a clean environment prior to install. This is not the case in the above scenario as `ocaml-findlib` has been previously installed and uninstalled - a process which does not delete `etc` files from the first install.

The change is to avoid duplicated dependent testing and defer dependent testing, where needed. So the above process would be like:

* Build `ocaml`.
* Test `ocaml` dependents, except those that depend on `ocaml-findlib`.
* Build `ocaml-findlib`.
* Test `ocaml-findlib` dependents (including build-time ones - we need to remember to cover them as `ocaml` would have done that as it's on the whitelist).

Example logs before and after this change in the scenario that `ocaml` and `ocaml-findlib` were modified in the same pull request:

<details>
<summary>Before</summary>

```
==> brew fetch --retry ocaml --build-bottle

==> brew uninstall --force ocaml

==> brew install --only-dependencies --verbose --build-bottle ocaml

==> brew install --verbose --build-bottle ocaml

==> brew bottle --verbose --json ocaml

==> brew linkage --test ocaml

==> brew install --only-dependencies --include-test ocaml

==> brew test ocaml --verbose

==> brew fetch --retry bibtex2html

==> brew install --build-from-source --only-dependencies bibtex2html

==> brew install --build-from-source bibtex2html

==> brew fetch --retry camlp5

==> brew install --build-from-source --only-dependencies camlp5

==> brew install --build-from-source camlp5

==> brew fetch --retry coccinelle

==> brew install --build-from-source --only-dependencies coccinelle

==> brew install --build-from-source coccinelle

==> brew fetch --retry comby

==> brew install --build-from-source --only-dependencies comby

==> brew install --build-from-source comby

==> brew fetch --retry compcert

==> brew install --build-from-source --only-dependencies compcert

==> brew install --build-from-source compcert

==> brew fetch --retry coq

==> brew install --build-from-source --only-dependencies coq

==> brew install --build-from-source coq

==> brew fetch --retry corectl

==> brew install --build-from-source --only-dependencies corectl

==> brew install --build-from-source corectl

==> brew fetch --retry dune

==> brew install --build-from-source --only-dependencies dune

==> brew install --build-from-source dune

==> brew fetch --retry flow

==> brew install --build-from-source --only-dependencies flow

==> brew install --build-from-source flow

==> brew fetch --retry haxe

==> brew install --build-from-source --only-dependencies haxe

==> brew install --build-from-source haxe

==> brew fetch --retry hevea

==> brew install --build-from-source --only-dependencies hevea

==> brew install --build-from-source hevea

==> brew fetch --retry hyperkit

==> brew install --build-from-source --only-dependencies hyperkit

==> brew install --build-from-source hyperkit

==> brew fetch --retry infer

==> brew install --build-from-source --only-dependencies infer

==> brew install --build-from-source infer

==> brew fetch --retry lablgtk

==> brew install --build-from-source --only-dependencies lablgtk

==> brew install --build-from-source lablgtk

==> brew fetch --retry ledit

==> brew install --build-from-source --only-dependencies ledit

==> brew install --build-from-source ledit

==> brew fetch --retry math-comp

==> brew install --build-from-source --only-dependencies math-comp

==> brew install --build-from-source math-comp

==> brew fetch --retry menhir

==> brew install --build-from-source --only-dependencies menhir

==> brew install --build-from-source menhir

==> brew install --only-dependencies ocaml-num

==> brew linkage --test ocaml-num

==> brew install --only-dependencies --include-test ocaml-num

==> brew test --verbose ocaml-num

==> brew uninstall --force ocaml-num

==> brew fetch --retry ocamlbuild

==> brew install --build-from-source --only-dependencies ocamlbuild

==> brew install --build-from-source ocamlbuild

==> brew fetch --retry ocamlsdl

==> brew install --build-from-source --only-dependencies ocamlsdl

==> brew install --build-from-source ocamlsdl

==> brew fetch --retry one-ml

==> brew install --build-from-source --only-dependencies one-ml

==> brew install --build-from-source one-ml

==> brew install --only-dependencies opam

==> brew linkage --test opam

==> brew install --only-dependencies --include-test opam

==> brew test --verbose opam

==> brew uninstall --force opam

==> brew fetch --retry ott

==> brew install --build-from-source --only-dependencies ott

==> brew install --build-from-source ott

==> brew fetch --retry pdfsandwich

==> brew install --build-from-source --only-dependencies pdfsandwich

==> brew install --build-from-source pdfsandwich

==> brew fetch --retry unison

==> brew install --build-from-source --only-dependencies unison

==> brew install --build-from-source unison

==> brew fetch --retry zero-install

==> brew install --build-from-source --only-dependencies zero-install

==> brew install --build-from-source zero-install

==> brew fetch --retry ocaml-findlib --build-bottle

==> brew install --only-dependencies --verbose --build-bottle ocaml-findlib

==> brew install --verbose --build-bottle ocaml-findlib

==> brew bottle --verbose --json ocaml-findlib

==> brew linkage --test ocaml-findlib

==> brew install --only-dependencies --include-test ocaml-findlib

==> brew test ocaml-findlib --verbose
```
</details>

<details>
<summary>After</summary>

```
==> brew fetch --retry ocaml --build-bottle

==> brew uninstall --force ocaml

==> brew install --only-dependencies --verbose --build-bottle ocaml

==> brew install --verbose --build-bottle ocaml

==> brew bottle --verbose --json ocaml

==> brew linkage --test ocaml

==> brew install --only-dependencies --include-test ocaml

==> brew test ocaml --verbose

==> brew fetch --retry bibtex2html

==> brew install --build-from-source --only-dependencies bibtex2html

==> brew install --build-from-source bibtex2html

==> brew fetch --retry camlp5

==> brew install --build-from-source --only-dependencies camlp5

==> brew install --build-from-source camlp5

==> brew fetch --retry coccinelle

==> brew install --build-from-source --only-dependencies coccinelle

==> brew install --build-from-source coccinelle

==> brew fetch --retry comby

==> brew install --build-from-source --only-dependencies comby

==> brew install --build-from-source comby

==> brew fetch --retry corectl

==> brew install --build-from-source --only-dependencies corectl

==> brew install --build-from-source corectl

==> brew fetch --retry dune

==> brew install --build-from-source --only-dependencies dune

==> brew install --build-from-source dune

==> brew fetch --retry flow

==> brew install --build-from-source --only-dependencies flow

==> brew install --build-from-source flow

==> brew fetch --retry haxe

==> brew install --build-from-source --only-dependencies haxe

==> brew install --build-from-source haxe

==> brew fetch --retry hevea

==> brew install --build-from-source --only-dependencies hevea

==> brew install --build-from-source hevea

==> brew fetch --retry hyperkit

==> brew install --build-from-source --only-dependencies hyperkit

==> brew install --build-from-source hyperkit

==> brew fetch --retry lablgtk

==> brew install --build-from-source --only-dependencies lablgtk

==> brew install --build-from-source lablgtk

==> brew fetch --retry ledit

==> brew install --build-from-source --only-dependencies ledit

==> brew install --build-from-source ledit

==> brew fetch --retry menhir

==> brew install --build-from-source --only-dependencies menhir

==> brew install --build-from-source menhir

==> brew fetch --retry ocamlbuild

==> brew install --build-from-source --only-dependencies ocamlbuild

==> brew install --build-from-source ocamlbuild

==> brew fetch --retry ocamlsdl

==> brew install --build-from-source --only-dependencies ocamlsdl

==> brew install --build-from-source ocamlsdl

==> brew fetch --retry one-ml

==> brew install --build-from-source --only-dependencies one-ml

==> brew install --build-from-source one-ml

==> brew install --only-dependencies opam

==> brew linkage --test opam

==> brew install --only-dependencies --include-test opam

==> brew test --verbose opam

==> brew uninstall --force opam

==> brew fetch --retry ott

==> brew install --build-from-source --only-dependencies ott

==> brew install --build-from-source ott

==> brew fetch --retry pdfsandwich

==> brew install --build-from-source --only-dependencies pdfsandwich

==> brew install --build-from-source pdfsandwich

==> brew fetch --retry unison

==> brew install --build-from-source --only-dependencies unison

==> brew install --build-from-source unison

==> brew fetch --retry zero-install

==> brew install --build-from-source --only-dependencies zero-install

==> brew install --build-from-source zero-install

==> brew fetch --retry ocaml-findlib --build-bottle

==> brew install --only-dependencies --verbose --build-bottle ocaml-findlib

==> brew install --verbose --build-bottle ocaml-findlib

==> brew bottle --verbose --json ocaml-findlib

==> brew linkage --test ocaml-findlib

==> brew install --only-dependencies --include-test ocaml-findlib

==> brew test ocaml-findlib --verbose

==> brew fetch --retry compcert

==> brew install --build-from-source --only-dependencies compcert

==> brew install --build-from-source compcert

==> brew fetch --retry coq

==> brew install --build-from-source --only-dependencies coq

==> brew install --build-from-source coq

==> brew fetch --retry infer

==> brew install --build-from-source --only-dependencies infer

==> brew install --build-from-source infer

==> brew fetch --retry math-comp

==> brew install --build-from-source --only-dependencies math-comp

==> brew install --build-from-source math-comp

==> brew install --only-dependencies ocaml-num

==> brew linkage --test ocaml-num

==> brew install --only-dependencies --include-test ocaml-num

==> brew test --verbose ocaml-num

==> brew uninstall --force ocaml-num
```
</details>

On the note of build time, some of the build time improvements could be quite large. For example, the libffi pull request (Homebrew/homebrew-core#51273) currently takes 31 hours to build, and that includes duplicating 2 hour builds of llvm, llvm@6 and llvm@7.

### DIFF
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -420,7 +420,7 @@ module Homebrew
         return
       end
 
-      build_dependents_from_source = %w[
+      build_dependents_from_source_whitelist = %w[
         cabal-install
         docbook-xsl
         ghc
@@ -428,30 +428,59 @@ module Homebrew
         ocaml
         openjdk
         rust
-      ].include?(formula_name)
+      ]
 
       uses_args = []
       uses_args << "--recursive" unless Homebrew.args.skip_recursive_dependents?
-      # whitelist specific formula where we want to test build dependants.
-      uses_args << "--include-build" if build_dependents_from_source
       dependents =
-        Utils.popen_read("brew", "uses", "--include-test", *uses_args, formula_name)
+        Utils.popen_read("brew", "uses", "--include-build", "--include-test", *uses_args, formula_name)
              .split("\n")
       dependents -= @formulae
       dependents = dependents.map { |d| Formulary.factory(d) }
 
-      if build_dependents_from_source
-        @source_dependents = dependents
-        @testable_dependents = @source_dependents.select(&:test_defined?)
-        @bottled_dependents = []
-        return
+      dependents = dependents.zip(dependents.map do |f|
+        if Homebrew.args.skip_recursive_dependents?
+          f.deps
+        else
+          f.recursive_dependencies
+        end.reject(&:optional?)
+      end)
+
+      # Defer formulae which could be tested later
+      # i.e. formulae that also depend on something else yet to be built in this test run.
+      dependents.select! do |_, deps|
+        still_to_test = @formulae - @formulae_that_have_been_built
+        (deps.map { |d| d.to_formula.full_name } & still_to_test).empty?
       end
 
-      @source_dependents = []
+      # Split into dependents that we could potentially be building from source and those
+      # we should not. The criteria is that it depends on a formula in the whitelist and
+      # that formula has been, or will be, built in this test run.
+      @source_dependents, dependents = dependents.partition do |_, deps|
+        deps.any? do |d|
+          full_name = d.to_formula.full_name
+
+          next false unless build_dependents_from_source_whitelist.include?(full_name)
+
+          @formulae.include?(full_name)
+        end
+      end
+
+      # From the non-source list, get rid of any dependents we are only a build dependency to
+      dependents.select! do |_, deps|
+        deps.reject { |d| d.build? && !d.test? }
+            .map(&:to_formula)
+            .include?(formula)
+      end
+
+      dependents = dependents.transpose.first.to_a
+      @source_dependents = @source_dependents.transpose.first.to_a
+
+      @testable_dependents = @source_dependents.select(&:test_defined?)
       @bottled_dependents = with_env(HOMEBREW_SKIP_OR_LATER_BOTTLES: "1") do
         dependents.select(&:bottled?)
       end
-      @testable_dependents = @bottled_dependents.select(&:test_defined?)
+      @testable_dependents += @bottled_dependents.select(&:test_defined?)
     end
 
     def unlink_conflicts(formula)
@@ -629,6 +658,7 @@ module Homebrew
       cleanup_during
 
       @category = "#{__method__}.#{formula_name}"
+      @formulae_that_have_been_built << formula_name
 
       formula = Formulary.factory(formula_name)
 
@@ -1071,6 +1101,7 @@ module Homebrew
     end
 
     def run
+      @formulae_that_have_been_built = []
       cleanup_before
       begin
         download
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -426,6 +426,8 @@ module Homebrew
         ghc
         go
         ocaml
+        ocaml-findlib
+        ocaml-num
         openjdk
         rust
       ]
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -780,6 +780,11 @@ module Homebrew
 
         @source_dependents.each do |dependent|
           install_dependent_from_source(dependent)
+
+          bottled = with_env(HOMEBREW_SKIP_OR_LATER_BOTTLES: "1") do
+            dependent.bottled?
+          end
+          install_bottled_dependent(dependent) if bottled
         end
         @bottled_dependents.each do |dependent|
           install_bottled_dependent(dependent)


Bo98:
The change in the build-from-source whitelist change does mean many more formulae could be built from source now. For example, a standalone ocaml-findlib update will now do so.

I see this as both a benefit and a downside depending on the formula. Perhaps I should think of something better, but I probably won't have the time tonight.

Bo98:
Code probably looks bad, sorry - I'll deal with any comments after my flight tomorrow.

lib/test.rb
```diff
@@ -354,7 +354,7 @@ def install_mercurial_if_needed(deps, reqs)
       end
     end
 
-    def setup_formulae_deps_instances(formula, formula_name)
+    def setup_formulae_deps_instances(formula, formula_name, still_to_test:)
```
MikeMcQuaid: Rather than passing these in I'd be tempted to use a `@formula_that_have_been_built` array or something which is shared between methods.

lib/test.rb
```diff
@@ -413,13 +413,19 @@ def setup_formulae_deps_instances(formula, formula_name)
...
         ocaml
+        ocaml-findlib
+        ocaml-num
         rust
-      ].include?(formula_name)
+      ]
+      build_dependents_from_source = build_dependents_from_source_whitelist.include?(formula_name) ||
+                                     (!(@formulae & build_dependents_from_source_whitelist).empty? &&
+                                      !(formula.recursive_dependencies.map { |d| d.to_formula.full_name } &
+                                        build_dependents_from_source_whitelist).empty?)
```
MikeMcQuaid: Would be good to split these into multiple variables (maybe even make the whole thing a function).

lib/test.rb
```diff
@@ -430,6 +436,16 @@ def setup_formulae_deps_instances(formula, formula_name)
              .split("\n")
       dependents -= @formulae
       dependents = dependents.map { |d| Formulary.factory(d) }
+      dependents = dependents.select do |f|
+        deps = if Homebrew.args.skip_recursive_dependents?
+          f.deps
+        else
+          f.recursive_dependencies
+        end
+        deps = deps.reject(&:optional?)
```
```suggestion
        deps.reject!(&:optional?)
```

Bo98:
The initial approach had some bad holes. For example, say if you threw cmake in among ocaml and ocaml-findlib (unlikely but there's probably better examples), then it could "defer" build-from-source tests to CMake but CMake won't actually run then because it is not on the whitelist.

It's a bit awkward to solve, but what I did was instead of checking the formula currently going through a build is on the whitelist, I test the dependents (which will search the recursive dependencies). We then split the dependents to ones to build from source and ones to install from bottles.

This means `--include-build` is always passed, so extra consideration is needed to filter those out again where needed.

I did notice one issue, though it is unrelated to these changes: installing dependents from source doesn't really properly check linkage since we never install from a bottle, so we don't know if revision bumps are required.

lib/test.rb
```diff
@@ -413,36 +435,60 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+      # Firstly, split into dependents that we could potentially be building from source
+      # and those we should not. The criteria is that it depends on a formula in the whitelist
+      # and that formula has been, or will be, built in this test run.
+      @source_dependents, dependents = dependents.partition do |f|
+        deps = if Homebrew.args.skip_recursive_dependents?
+          f.deps
+        else
+          f.recursive_dependencies
+        end
+        deps.reject!(&:optional?)
```
MikeMcQuaid: Looks like a bunch of logic above here could be used in a way that doesn't require their reputation below?

Bo98:
> I'm also wondering if we don't actively want to try and test both the bottle and building from source in some of these situations?

I feel we should in order to achieve full coverage. We don't currently for those on the whitelist, but as a result we are missing out on potential linkage failures in existing bottles.

lib/test.rb
Bo98: Yeah, I feel so. I suppose I could maintain a list of dependencies of the dependents (or zip the two together).

MikeMcQuaid:
> Yes, testing both source builds and bottle installs will take more time on build from source runs - but equally we are making time savings here eliminating duplicated builds when such formulae are revision bumped.

Cool, I'm up for that, then. Should make things a bit easier codewise as just need to skip installing from bottle/from source if you've already done that particular combination.

Bo98:
I've been busy the last week but I've managed to tweak it a little just now.

lib/test.rb
```diff
@@ -420,38 +444,58 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+      # Firstly, split into dependents that we could potentially be building from source
+      # and those we should not. The criteria is that it depends on a formula in the whitelist
+      # and that formula has been, or will be, built in this test run.
+      @source_dependents, dependents = dependents.partition do |_, deps|
+        deps.any? do |d|
+          dep_formula = d.to_formula
+
+          next false unless @formulae.include?(dep_formula.full_name)
+
+          build_dependents_from_source?(dep_formula)
```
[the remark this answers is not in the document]
Bo98: Indeed - I think the repetition was a consequence of iterative changes. I'll simplify this.

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
-      ].include?(formula_name)
+      ]
 
       uses_args = []
       uses_args << "--recursive" unless Homebrew.args.skip_recursive_dependents?
-      # whitelist specific formula where we want to test build dependants.
-      uses_args << "--include-build" if build_dependents_from_source
       dependents =
-        Utils.popen_read("brew", "uses", "--include-test", *uses_args, formula_name)
+        Utils.popen_read("brew", "uses", "--include-build", "--include-test", *uses_args, formula_name)
```
Bo98: We cannot take this simply from a top level view. We need to consider it from a dependent level.

Why? Consider this example:

A is on the whitelist and builds dependent B
C is not on the whitelist but also wants to test dependent B.

B should be deferred (skipped) from the A build, but when C builds dependents it should still build B from source - something that wouldn't be done if we only consider it on the top level of A & C.

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
              .split("\n")
       dependents -= @formulae
       dependents = dependents.map { |d| Formulary.factory(d) }
 
-      if build_dependents_from_source
-        @source_dependents = dependents
-        @testable_dependents = @source_dependents.select(&:test_defined?)
-        @bottled_dependents = []
-        return
```
Bo98: In light of the nature where some dependents will be built from source and some not, this logic has been replaced (see below).

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+      dependents = dependents.zip(dependents.map do |f|
+        if Homebrew.args.skip_recursive_dependents?
+          f.deps
+        else
+          f.recursive_dependencies
+        end.reject(&:optional?)
+      end)
```
Bo98: We check the dependencies a few times. This just avoids code duplication.

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+        deps.any? do |d|
+          full_name = d.to_formula.full_name
+
+          next false unless build_dependents_from_source_whitelist.include?(full_name)
+
+          @formulae.include?(full_name)
+        end
       end
 
-      @source_dependents = []
```
Bo98: We look at the dependents and split them into those that should be built from source and which should not. The check is to see if any of the dependent's dependencies (recursively if enabled) is on the whitelist _and_ that dependency is in the list of top-level formulae being tested.

This should be seen as a replacement of the existing build from source selection logic rather than just an addition. It allows much greater flexibility to defer builds and thus avoid repeating builds or polluting `etc` and `var`.

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+          f.recursive_dependencies
+        end.reject(&:optional?)
+      end)
+
+      # Defer formulae which could be tested later
+      # i.e. formulae that also depend on something else yet to be built in this test run.
+      dependents.select! do |_, deps|
+        still_to_test = @formulae - @formulae_that_have_been_built
+        (deps.map { |d| d.to_formula.full_name } & still_to_test).empty?
       end
```
Bo98: This is the defer logic. If a dependent is due to be built from source, and is set to be tested again in the future (determined by checking that dependent's dependencies), it is deferred.

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+      # that formula has been, or will be, built in this test run.
+      @source_dependents, dependents = dependents.partition do |_, deps|
+        deps.any? do |d|
+          full_name = d.to_formula.full_name
+
+          next false unless build_dependents_from_source_whitelist.include?(full_name)
+
+          @formulae.include?(full_name)
+        end
+      end
```
Bo98: We look at the dependents and split them into those that should be built from source and which should not. The check is to see if any of the dependent's dependencies (recursively if enabled) is on the whitelist _and_ that dependency is in the list of top-level formulae being tested.

This should be seen as a replacement of the existing build from source selection logic rather than just an addition. It allows much greater flexibility to defer builds and thus avoid repeating builds or polluting `etc` and `var`.

Otherwise, deferring would not be possible for mixed BFS & bottle cases and additional checks would need to be done at the defer stage to prevent that.

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+
+      # From the non-source list, get rid of any dependents we are only a build dependency to
+      dependents.select! do |_, deps|
+        deps.reject { |rd| rd.build? && !rd.test? }
+            .map(&:to_formula)
+            .include?(formula)
+      end
```
Bo98: This is a consequence of always having `--include-build`. We are not interested in testing dependents that are just there via `:build` and we are not building from source (as per the whitelist).

I suppose this could actually be expanded to cover `bottle :unneeded` but it didn't before so is unchanged in that regard.

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+
+      dependents = dependents.transpose.first.to_a
+      @source_dependents = @source_dependents.transpose.first.to_a
```
Bo98: Basically undoing the zip we did earlier (443-449).

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+
+      @testable_dependents = @source_dependents.select(&:test_defined?)
       @bottled_dependents = with_env(HOMEBREW_SKIP_OR_LATER_BOTTLES: "1") do
         dependents.select(&:bottled?)
       end
-      @testable_dependents = @bottled_dependents.select(&:test_defined?)
+      @testable_dependents += @bottled_dependents.select(&:test_defined?)
```
Bo98: There is effectively no change here.

lib/test.rb
```diff
@@ -629,6 +660,7 @@ def formula(formula_name)
       cleanup_during
 
       @category = "#{__method__}.#{formula_name}"
+      @formulae_that_have_been_built << formula_name
```
Bo98: This is used to detect which formulae are still left to build.

lib/test.rb
```diff
@@ -748,6 +780,11 @@ def formula(formula_name)
 
         @source_dependents.each do |dependent|
           install_dependent_from_source(dependent)
+
+          bottled = with_env(HOMEBREW_SKIP_OR_LATER_BOTTLES: "1") do
+            dependent.bottled?
+          end
+          install_bottled_dependent(dependent) if bottled
```
Bo98: This is a separate change, and a separate commit. But one it makes sense to do at the same time to ensure test coverage remains high.

When we build from source, we don't actually properly test linkage of existing bottles. For example, if OCaml is updated, the tests will currently never detect that ocaml-findlib needs a revision bump because the existing bottles are never tested. This change make it run _both_ a build-from-source and a bottle install test.

lib/test.rb
```diff
@@ -1071,6 +1108,7 @@ def devel_only_tap?(formula)
     end
 
     def run
+      @formulae_that_have_been_built = []
```
Bo98: This makes sense to be in `run` so you don't have to recreate the test in order to re-run (though that never happens anyway).

Bo98:
> what seems like a bit of an edge case

To clarify on this note, it depends on the angle you view it:

* On the note of improving build time, no I don't see this as just an edge case. This will be beneficial to a fair number of pull requests.
* On the note of producing bottles that are not broken out-of-the-box, yes that is probably an edge case. I've only seen such a broken bottle once. But it's something that shouldn't really ever be happening.

lib/test.rb
```diff
@@ -420,38 +420,69 @@ def setup_formulae_deps_instances(formula, formula_name)
...
+
+          next false unless build_dependents_from_source_whitelist.include?(full_name)
+
+          @formulae.include?(full_name)
+        end
+      end
+
+      # From the non-source list, get rid of any dependents we are only a build dependency to
+      dependents.select! do |_, deps|
+        deps.reject { |rd| rd.build? && !rd.test? }
```
MikeMcQuaid: Why `rd` here? Could this be `d` or `dep`?

Bo98: That'll be because there used to be a `d` it shadowed before I refactored it a little. Will fix!

Indeed I do have permissions.

Bo98:
Just some results here:

Before this change, a libffi pull request took 31 hours to build.
After this change, libffi takes 16-17 hours to build and that includes one more LLVM formula than before (2 hours).

PR #341, opened 2020-03-19
test: install subversion if needed

dawidd6:
Should prevent situations like this (on Linux):

```
==> brew fetch --retry netpbm --build-bottle --force
==> Cloning https://svn.code.sf.net/p/netpbm/code/stable
==> Checking out 3750
You must: brew install svn
==> Retrying download
==> Cloning https://svn.code.sf.net/p/netpbm/code/stable
==> Checking out 3750
You must: brew install svn
Error: Failed to download resource "netpbm"
Failure while executing; `svn checkout https://svn.code.sf.net/p/netpbm/code/stable /github/home/.cache/Homebrew/netpbm--svn --quiet -r 3750` exited with 1. Here's the output:
You must: brew install svn


==> FAILED
```

https://github.com/Homebrew/linuxbrew-core/pull/19893/checks?check_run_id=518562926

### DIFF
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -354,6 +354,13 @@ module Homebrew
       end
     end
 
+    def install_subversion_if_needed(deps, reqs)
+      if (deps | reqs).any? { |d| d.name == "subversion" && d.build? }
+        test "brew", "install", "subversion",
+             env: { "HOMEBREW_DEVELOPER" => nil }
+      end
+    end
+
     def setup_formulae_deps_instances(formula, formula_name)
       conflicts = formula.conflicts
       formula.recursive_dependencies.each do |dependency|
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @dawidd6!

iMichka:
I still see this error: https://github.com/Homebrew/linuxbrew-core/runs/1114338360?check_suite_focus=true

Should it be "svn" instead of "subversion" ?

dawidd6:
It worked before just fine.

> Should it be "svn" instead of "subversion" ?
https://github.com/Homebrew/brew/blob/7589a350ccf76d9a75cb76dd6db9222f17b67701/Library/Homebrew/dependency_collector.rb#L65

iMichka:
The weird thing is that we have a netpbm bottle, but it tries to install from source nevertheless.

PR #342, opened 2020-03-20
test: build openjdk dependents from source

jonchang:
Request of https://github.com/Homebrew/homebrew-core/pull/51883#issuecomment-601579482

This pull request will build formulae that depend on `openjdk` from source on test-bot runs.


MikeMcQuaid:
Makes sense, thanks @jonchang!

PR #345, opened 2020-03-23
test: add docbook-xsl to BFS whitelist

Bo98:
I had hoped to get this before Homebrew/homebrew-core#51747 was merged, so let's at least try to get this in before the subsequent fixup pull requests are also merged, and restart any already-run CI jobs as needed.


MikeMcQuaid:
Thanks again @Bo98!

PR #346, opened 2020-03-28
Fix style

Bo98:
Because of [this](https://github.com/Homebrew/brew/blob/1a538028cd05e99630ebe7204f688deeb3e8cbaf/Library/Homebrew/style.rb#L60-L61) and [this](https://github.com/Homebrew/brew/blob/708010e6fd07a9a606b12c0cc735ac0706bd8d9e/Library/.rubocop_rspec.yml#L4-L5), `brew style homebrew/test-bot` only runs on two files in this repository.

This should be fixed over in Homebrew/brew at some point. I've fixed the current style violations here.


MikeMcQuaid:
Thanks again @Bo98!

PR #347, opened 2020-03-29
test-bot: teach --publish to obey --dry-run

jonchang:
This prevents erroneous bottle publication when we are merely seeing what commands will be run.


MikeMcQuaid:
Good call, thanks @jonchang!

PR #350, opened 2020-04-01
lib/test: skip disabled/deprecated formula.

MikeMcQuaid:
Part of https://github.com/Homebrew/brew/issues/7029.

This will allow us to make deprecated/disabled formulae not cause CI issues.

Requires https://github.com/Homebrew/brew/pull/7190 to be merged to work.

### DIFF
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -578,6 +578,13 @@ module Homebrew
     def install_dependent_from_source(dependent)
       return if Homebrew.args.fast? || !satisfied_requirements?(dependent, :stable)
 
+      if dependent.deprecated? || dependent.disabled?
+        verb = dependent.deprecated? ? :deprecated : :disabled
+        puts "#{dependent.full_name} has been #{verb}!"
+        skip dependent.name
+        return
+      end
+
       cleanup_during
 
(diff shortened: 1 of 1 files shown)


lib/test.rb
```diff
@@ -657,6 +671,12 @@ def formula(formula_name)
       @formulae_that_have_been_built << formula_name
 
       formula = Formulary.factory(formula_name)
+      if formula.deprecated? || formula.disabled?
+        verb = formula.deprecated? ? :deprecated : :disabled
+        puts "#{formula.full_name} has been #{verb}!"
+        skip formula.name
+        return
+      end
```
Bo98, line 679: It makes sense to skip as a dependent but do we actually want to skip it if changes to that formula are the main focus of the pull request being tested?

MikeMcQuaid: I think so. Definitely if they are disabled but even when deprecated I wonder what the merits are of continuing to run CI jobs on them. CC @sjackman @SMillerDev @reitermarkus for thoughts

MikeMcQuaid: @Bo98 Let me know what you think of this now.

PR #355, opened 2020-04-19
Delete unused code.

MikeMcQuaid:
Now that we've decommissioned jenkins.brew.sh we can delete all Jenkins related code. While doing this, audit remaining code to ensure that all methods, parameters, environment variables and configurations are those used and supported by Homebrew.

### DIFF
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -102,14 +102,9 @@ module Homebrew
         @git, "-C", @repository, "symbolic-ref", "HEAD"
       ).gsub("refs/heads/", "").strip
 
-      # Use Jenkins GitHub Pull Request Builder variables for pull request jobs.
-      if ENV["ghprbPullLink"]
-        @url = ENV["ghprbPullLink"]
-        @hash = nil
-        test @git, "-C", @repository, "checkout", "origin/master"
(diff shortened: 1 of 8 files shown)


lib/test.rb
```diff
@@ -1124,7 +994,6 @@ def run
       ensure
         cleanup_after
       end
-      check_results
```
Bo98, line 1127: This line needs to stay. `run` should return a boolean that indicates whether there were no failures.

lib/test.rb
```diff
@@ -725,40 +684,13 @@ def formula(formula_name)
...
-
-      install_args += shared_install_args
+      install_args = ["--verbose"]
+      install_args << "--build-bottle" unless formula.bottle_disabled?
       install_args << formula_name
 
       # Don't care about e.g. bottle failures for dependencies.
       install_passed = false
-      if !Homebrew.args.fast? || formula_bottled || formula.bottle_unneeded?
+      if formula.bottled? || formula.bottle_unneeded?
```
Bo98, line 693: This is a logic change. Previously this would always be true if `--fast` wasn't set. With this change, that is no longer the case.

This condition now evaluates to false for new formulae - they don't have a bottle block.

PR #356, opened 2020-04-19
Revert "Delete unused code."

iMichka:
Reverts Homebrew/homebrew-test-bot#355

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -52,5 +52,6 @@ jobs:
     - name: Run brew test-bot --dry-run tests
       if: matrix.os == 'macOS-latest'
       run: |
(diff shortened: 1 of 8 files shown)


iMichka:
`--ci-upload` is still used by linuxbrew-core to upload bottles. Until we find an alternative, I would like to keep this around. As the commit was not atomic and multiple commands have been removed at once, I can't really easily extract only the `--ci-upload` part.

We have multiple bottle uploads waiting in linuxbrew-core right now, so we need to revert this ASAP. As this is test-bot code and we have the habit to revert broken stuff, I am doing what we often do here: revert :)

We should first split this PR in at least two parts: one to remove jenkins code, and maybe another one to remove `--ci-upload` when not needed anymore.

PR #357, opened 2020-04-20
Delete unused code.

MikeMcQuaid:
Take #2 of https://github.com/Homebrew/homebrew-test-bot/pull/355 but with `--ci-upload` non-Jenkins bits kept around and the omissions pointed out by @Bo98.

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -14,24 +14,14 @@ module Homebrew
 
       switch "--dry-run",
              description: "print what would be done rather than doing it."
-      switch "--keep-logs",
-             description: "write and keep log files under `./brewbot/`."
       switch "--cleanup",
              description: "clean all state from the Homebrew directory. Use with care!"
       switch "--skip-setup",
              description: "don't check if the local system is set up correctly."
-      switch "--junit",
-             description: "generate a JUnit XML test results file."
-      switch "--no-bottle",
-             description: "run `brew install` without `--build-bottle`."
       switch "--keep-old",
              description: "run `brew bottle --keep-old` to build new bottles for a single platform."
       switch "--skip-relocation",
              description: "run `brew bottle --skip-relocation` to build new bottles that don't require relocation."
-      switch "--HEAD",
-             description: "run `brew install` with `--HEAD`."
-      switch "--local",
-             description: "ask Homebrew to write verbose logs under `./logs/` and set `$HOME` to `./home/`"
       flag   "--tap=",
              description: "use the `git` repository of the given tap."
       switch "--fail-fast",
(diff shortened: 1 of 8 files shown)


cmd/test-bot.rb
```diff
@@ -83,40 +62,21 @@ def test_bot
...
 
   def setup_argv_and_env
-    jenkins = !ENV["JENKINS_HOME"].nil?
-    github_actions_self_hosted = !ENV["GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED"].nil?
-
-    github_actions = !ENV["GITHUB_ACTIONS"].nil?
-    if github_actions
-      ARGV << "--verbose" << "--ci-auto" << "--no-pull"
+    github_actions = ENV["GITHUB_ACTIONS"].present?
+    if ENV["GITHUB_ACTIONS"].present?
```
Bo98, line 66:
```suggestion
    github_actions = ENV["GITHUB_ACTIONS"].present?
    if github_actions
```

MikeMcQuaid: Done, thanks ❤️

lib/test.rb
```diff
@@ -841,15 +737,12 @@ def deleted_formula(formula_name)
     def readall
       @category = __method__
       return if @skip_homebrew
```
Bo98, line 739: Is this set anywhere?

MikeMcQuaid: Nope, removed!

cmd/test-bot.rb
```diff
@@ -56,12 +42,12 @@ def test_bot_args
...
-             description: "use the Homebrew testing CI options. Implies `--cleanup`: use with care!"
-      switch "--ci-auto",
-             description: "automatically pick one of the Homebrew CI options based on the environment. "\
-                          "Implies `--cleanup`: use with care!"
+              description: "use the Homebrew pull request CI options. Implies `--cleanup`: use with care!"
+switch "--ci-testing",
+              description: "use the Homebrew testing CI options. Implies `--cleanup`: use with care!"
+       switch "--ci-auto",
+              description: "automatically pick one of the Homebrew CI options based on the environment. "\
+                           "Implies `--cleanup`: use with care!"
```
Bo98, line 50: Funky indentation.

MikeMcQuaid: Left in by mistake.

MikeMcQuaid:
> Just for the sake of covering everything and seeing where the removed options are used, I went through some taps via searching on GitHub:

Thanks for doing that 🙇 . The annoying thing is we don't really have a deprecation path for these. I'm going to just remove some of these options and re-add if necessary/complaints arise.

> * The `--ci-*` options are very commonly used on Travis.

They are pretty unnecessary at this point given they were only running `--cleanup`. Furthermore, all Travis special-cased code has been removed long ago meaning `brew doctor` should fail pretty often.

> * A few lower-profile taps used `--no-bottle`. Some abandoned.
> * osrf/homebrew-simulation use `--fast`.
> * I saw one tap use `--HEAD`, forgot which one.
> though they also use `--skip-homebrew` which I don't see about.
> For the devel/HEAD-only stuff: it's possibly used in taps but I don't have numbers on that.

We haven't used these for a long time and they complicate the code (unnecessarily in my opinion).

> * Autobrew uses `--local`, 

`--local` is now default behaviour.

---

I see `brew test-bot` being killed in the medium term and replaced with a collection of scripts and GitHub Actions that people can piece together if they wish.

PR #363, opened 2020-04-22
test-bot: cleanup parameters, variables.

MikeMcQuaid:
Cleanup all use of parameters and instance variables to keep data more closely scoped to existing functions (where possible).

This will enable future PRs to split up the `test-bot` output.

### DIFF
--- a/lib/test.rb
+++ b/lib/test.rb
@@ -22,41 +22,22 @@ module Homebrew
     attr_reader :log_root, :category, :name, :steps
 
     def initialize(argument, tap:, git:, skip_setup: false, skip_cleanup_before: false, skip_cleanup_after: false)
-      @hash = nil
-      @url = nil
-      @formulae = []
-      @added_formulae = []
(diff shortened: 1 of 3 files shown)


Bo98:
`@formulae_that_have_been_built` in `run` still refers to the old name.

lib/test.rb
```diff
@@ -136,51 +136,53 @@ def download
...
+        diff_start_sha1 = "#{hash}^"
+        diff_end_sha1 = hash
+        name = hash
       # Handle a URL being passed on the command-line e.g.
       #   brew test-bot https://github.com/Homebrew/homebrew-core/pull/678
-      elsif @url
+      elsif url
+        diff_start_sha1 = current_sha1
         unless Homebrew.args.no_pull?
-          diff_start_sha1 = current_sha1
```
Bo98, line 166: Possibly problematic moving this? GitHub Actions set `--no-pull`.

PR #369, opened 2020-04-22
cmd/test-bot: cleanup by default on GitHub Actions.

MikeMcQuaid:
The default VMs ship with some stuff that makes `brew doctor` sad.


sjackman:
Thanks, Mike!

PR #370, opened 2020-04-22
test_bot: restrict CLI arguments.

MikeMcQuaid:
Accept only formula, `HEAD` or GitHub Actions commits for building.

I believe other functionality is unused. If I'm wrong: correct me.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -63,5 +63,5 @@ jobs:
       if: matrix.os == 'macOS-latest'
       run: |
         brew test-bot --ci-upload --dry-run
-        brew test-bot https://github.com/Homebrew/brew/pull/4170 --dry-run
-        brew test-bot https://github.com/Homebrew/brew/commit/dc96e6f73551fbadd0c2169c3a79c47a936f71ae --dry-run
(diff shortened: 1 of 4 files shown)


scpeters:
we have been passing a URL to test-bot to build bottles in our tap, but I'll try to look into the most recent changes to see if we can do it differently

MikeMcQuaid:
@scpeters Sorry to hear that. Your best bet will be trying to set the `diff_start_sha1` and `diff_end_sha1` some other way (perhaps by faking the GitHub Actions environment variables).

What CI provider are you using? I'd recommend a move to GitHub Actions for the easiest `test-bot` usage. Give me a shout in Slack if you want to talk more about this.

scpeters:
we have our own custom jenkins instance

I will indeed look into using GitHub actions instead and reach out if I have questions

MikeMcQuaid:
> we have our own custom jenkins instance

👍 cool. 



> I will indeed look into using GitHub actions instead

Yeh, we found it one less server to run. You can probably steal a bunch from homebrew-core's workflow files at this point that are running on our self-hosted Actions workers.

PR #375, opened 2020-04-23
lib: cleanup output.

MikeMcQuaid:
- Don't use `/usr/bin/git` when `git` will do
- Output what method is being run each time

### DIFF
--- a/lib/step.rb
+++ b/lib/step.rb
@@ -21,6 +21,7 @@ module Homebrew
              .join(" ")
              .gsub("#{HOMEBREW_LIBRARY}/Taps/", "")
              .gsub("#{HOMEBREW_PREFIX}/", "")
+             .gsub(%r{^/usr/bin/}, "")
     end
 
     def passed?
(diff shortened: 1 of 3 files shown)


lib/step.rb
```diff
@@ -21,6 +21,7 @@ def command_trimmed
              .join(" ")
              .gsub("#{HOMEBREW_LIBRARY}/Taps/", "")
              .gsub("#{HOMEBREW_PREFIX}/", "")
+             .gsub("/usr/bin/", "/")
```
Bo98, line 24: Wouldn't that replace `/usr/bin/git` with `/git`?

Bo98: How does this deal with something say in `/Library/Developer/CommandLineTools/usr/bin` (though maybe there's nothing that would do that though that needs worrying about)

MikeMcQuaid: Yes, yes it would 😁. Fixed!

MikeMcQuaid: > (though maybe there's nothing that would do that that needs worrying about)

there isn't but handled anyway now

PR #376, opened 2020-04-23
Make GitHub Actions output less verbose

MikeMcQuaid:
- Don't pass `--verbose` by default with GitHub Actions (instead, just output failed steps)
- Always output `brew config` even when it doesn't fail (as it almost never does).
- Don't run `brew --env` as we don't really care about the output.
- Output repository revisions with a headline and after they won't change again.
- More consistent spacing and colours.


dawidd6:
I would propose some other color like yellow, rather than red. Red just feels like error all the time.

MikeMcQuaid:
@dawidd6 Good idea. I have avoided green and red and used yellow for a warning.

PR #377, opened 2020-04-26
Provide arguments to only test one step.

MikeMcQuaid:
This will allow the test-bot logs to be more segmented in GitHub Actions and avoid wasting time testing some formulae when others have invalid syntax.


Bo98:
Won't `steps_output.txt` get deleted on every run and thus only reflect the results of the last `brew test-bot`?

MikeMcQuaid:
@Bo98 Yup. We only really care about it for formulae, though.

PR #399, opened 2020-05-14
tests/formulae: add erlang to BFS whitelist

bayandin:
This PR adds erlang to build-from-source whitelist. This should help to catch build problems on erlang-dependant formulae earlier and help to decide if we need to add a versioned formula for the previous version (in case of major updates).


MikeMcQuaid:
Thanks @bayandin!

Bo98:
:+1:

PR #436, opened 2020-07-06
lib/test_cleanup: more rigorous and reliable file deleting

Bo98:
* Ensure permissions are correct to ensure deletion is a success.
* Cover hidden files & directories (so that we can also check the permissions of them).

Permission fixing is one pass while deletion is second pass. This is so we don't try to delete a parent directory before we've managed to fix the permissions of its children, as that will fail.


MikeMcQuaid:
Looks good, nice work!

PR #446, opened 2020-07-29
formulae: adjust testing behaviour.

MikeMcQuaid:
- when testing a formula as a dependent of another: retry the test on failures (requires https://github.com/Homebrew/brew/pull/8135)
- when testing a added/modified formula:
    - don't retry the test (we want to know if it's flaky)
    - don't try to test dependents if the test fails

This should improve both the reliability and performance of PRs with large numbers of dependents.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -491,7 +491,7 @@ module Homebrew
         if @testable_dependents.include? dependent
           test "brew", "install", "--only-dependencies", "--include-test",
                                   dependent.full_name
-          test "brew", "test", "--verbose", dependent.full_name
+          test "brew", "test", "--retry", "--verbose", dependent.full_name
         end
 
         test "brew", "uninstall", "--force", dependent.full_name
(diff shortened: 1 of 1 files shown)


lib/tests/formulae.rb
```diff
@@ -607,22 +607,30 @@ def formula!(formula_name)
...
 
-          if formula.test_defined?
+          test_passed_or_missing = if formula.test_defined?
             test "brew", "install", "--only-dependencies", "--include-test",
                                     formula_name
+            # Intentionally not passing --retry here to avoid papering over
+            # flaky tests when a formula isn't being pulled in as a dependent.
             test "brew", "test", "--verbose", formula_name
+
+            steps.last.passed?
```
Bo98, line 617: Considering it is called `test_passed_or_missing`, don't we want:

```suggestion
            steps.last.passed?
          else
            true
```

MikeMcQuaid: Yes, it was there in a previous iteration, I swear 😅.

MikeMcQuaid: https://github.com/Homebrew/homebrew-test-bot/pull/447

PR #452, opened 2020-07-30
Remove global flags.

reitermarkus:


cmd/test-bot.rb
```diff
@@ -61,7 +61,6 @@ def test_bot_args
              description: "Only run the post-cleanup step. Needs `--cleanup`."
       conflicts "--only-cleanup-before", "--only-setup", "--only-tap-syntax",
                 "--only-formulae", "--only-cleanup-after"
-      switch :debug
```
MikeMcQuaid, line 64: This is definitely undesirable. All commands should support `--debug` because they print out a full bracktrace and are extremely useful for debugging (both external commands and Homebrew/brew). Could you instead figure out a way it perhaps doesn't need to be specified but is always accepted (if that's not what's already done).

reitermarkus: > Could you instead figure out a way it perhaps doesn't need to be specified but is always accepted

I think it already works like that now, and it definitely should work like that after https://github.com/Homebrew/brew/pull/8144 is merged.

MikeMcQuaid: Cool, thanks!

PR #453, opened 2020-07-30
lib/tests/formulae: check GITHUB_REPOSITORY

Bo98:
If `GITHUB_REPOSITORY` is not the same as the tap being used, then we should not be considering `GITHUB_SHA` etc.

This is the case in Homebrew/brew, but could in theory be elsewhere through the use of `--tap`.

Also revert to `popen_read` for `rev_parse` since it seems to be intentionally designed in a way that it returns nothing if it can't find the revision.

Technically, this should probably check `remote` rather than `full_name`, in case someone is using a non-conventional repository name.

(The split into a separate function is to satisfy Rubocop)

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -42,7 +42,7 @@ module Homebrew
       end
 
       def rev_parse(ref)
-        Utils.safe_popen_read(git, "-C", repository, "rev-parse", "--verify", ref).strip
+        Utils.popen_read(git, "-C", repository, "rev-parse", "--verify", ref).strip
       end
 
       def current_sha1
@@ -82,27 +82,31 @@ module Homebrew
             "#{@argument} is not detected from GitHub Actions or a formula name!"
         end
 
-        # Use GitHub Actions variables for pull request jobs.
-        if ENV["GITHUB_BASE_REF"].present? && ENV["GITHUB_SHA"].present?
-          diff_start_sha1 = rev_parse("origin/#{ENV["GITHUB_BASE_REF"]}")
-          if diff_start_sha1.blank?
-            test git, "-C", repository, "fetch", "--depth=1",
-                 "origin", "+refs/heads/#{ENV["GITHUB_BASE_REF"]}",
-                 args: args
-            diff_start_sha1 = rev_parse("origin/#{ENV["GITHUB_BASE_REF"]}")
-          end
-          diff_end_sha1 = ENV["GITHUB_SHA"]
-        # Use GitHub Actions variables for branch jobs.
-        elsif ENV["GITHUB_SHA"].present?
-          diff_end_sha1 = diff_start_sha1 = ENV["GITHUB_SHA"]
-        # Otherwise just use the current SHA-1 (which may be overriden later)
-        else
-          if ENV["CI"] || ENV["GITHUB_ACTIONS"]
+        if ENV["GITHUB_REPOSITORY"].blank? || ENV["GITHUB_SHA"].blank?
+          if ENV["GITHUB_ACTIONS"]
+            odie <<~EOS
+              We cannot find the needed GitHub Actions environment variables! Check you have e.g. exported them to a Docker container.
+            EOS
+          elsif ENV["CI"]
             onoe <<~EOS
-              No known CI provider detected! If you are using GitHub Actions then we cannot find the expected  environment variables! Check you have e.g. exported them to a Docker container.
+              No known CI provider detected! If you are using GitHub Actions then we cannot find the expected environment variables! Check you have e.g. exported them to a Docker container.
             EOS
           end
-          diff_end_sha1 = diff_start_sha1 = current_sha1
+        elsif tap.present? && tap.full_name.casecmp(ENV["GITHUB_REPOSITORY"]).zero?
+          # Use GitHub Actions variables for pull request jobs.
+          if ENV["GITHUB_BASE_REF"].present?
+            diff_start_sha1 = rev_parse("origin/#{ENV["GITHUB_BASE_REF"]}")
+            if diff_start_sha1.blank?
+              test git, "-C", repository, "fetch", "--depth=1",
+                   "origin", "+refs/heads/#{ENV["GITHUB_BASE_REF"]}",
+                   args: args
+              diff_start_sha1 = rev_parse("origin/#{ENV["GITHUB_BASE_REF"]}")
+            end
+            diff_end_sha1 = ENV["GITHUB_SHA"]
+          # Use GitHub Actions variables for branch jobs.
+          else
+            diff_end_sha1 = diff_start_sha1 = ENV["GITHUB_SHA"]
+          end
         end
 
         if diff_start_sha1.present? && diff_end_sha1.present?


lib/tests/formulae.rb
```diff
@@ -64,6 +64,27 @@ def diff_formulae(start_revision, end_revision, path, filter)
         end.compact
       end
 
+      def github_actions_revisions
```
MikeMcQuaid, line 67: Can you line this? The indirection makes it harder to follow when it's only called once and it's harder to see what the changes you've made are.

MikeMcQuaid:
> (The split into a separate function is to satisfy Rubocop)

What is the failure here?

lib/tests/formulae.rb
```diff
@@ -82,39 +103,27 @@ def detect_formulae!(args:)
...
-            EOS
-          end
-          diff_end_sha1 = diff_start_sha1 = current_sha1
+        elsif ENV["CI"]
+          onoe <<~EOS
+            No known CI provider detected! If you are using GitHub Actions then we cannot find the expected environment variables! Check you have e.g. exported them to a Docker container.
+          EOS
         end
 
+        diff_start_sha1 = current_sha1 if diff_start_sha1.blank?
```
MikeMcQuaid, line 117: Why is this moved? I don't think it should be here because we don't want to calculate the merge base based on these.

Bo98: The behaviour isn't really too different than before - if `GITHUB_SHA` wasn't present then it already did `diff_start_sha1 = diff_end_sha1 = current_sha1`. This was just a replacement of that line but perhaps it was just better to just remove the line and leave this below.

lib/tests/formulae.rb
Bo98: It was about too many if blocks. I perhaps rushed it a little - I'll think of something better.

lib/tests/formulae.rb
```diff
@@ -82,39 +103,27 @@ def detect_formulae!(args:)
...
-        if ENV["GITHUB_BASE_REF"].present? && ENV["GITHUB_SHA"].present?
-          diff_start_sha1 = rev_parse("origin/#{ENV["GITHUB_BASE_REF"]}")
-          if diff_start_sha1.blank?
-            test git, "-C", repository, "fetch", "--depth=1",
-                 "origin", "+refs/heads/#{ENV["GITHUB_BASE_REF"]}",
-                 args: args
-            diff_start_sha1 = rev_parse("origin/#{ENV["GITHUB_BASE_REF"]}")
+        if ENV["GITHUB_REPOSITORY"].present?
+          # We're using GitHub Actions. Check the repo matches the tap we are using.
+          if tap&.full_name&.downcase == ENV["GITHUB_REPOSITORY"].downcase
```
MikeMcQuaid, line 108:
```suggestion
        if tap.present? && ENV["GITHUB_REPOSITORY"].present? && tap.full_name.downcase == ENV["GITHUB_REPOSITORY"].downcase
```

Would that work?

lib/tests/formulae.rb
MikeMcQuaid: Yeh, would be ideal to scope that to only if `GITHUB_SHA` isn't present.

lib/tests/formulae.rb
```diff
@@ -82,39 +103,27 @@ def detect_formulae!(args:)
...
-          diff_end_sha1 = diff_start_sha1 = ENV["GITHUB_SHA"]
-        # Otherwise just use the current SHA-1 (which may be overriden later)
-        else
-          if ENV["CI"] || ENV["GITHUB_ACTIONS"]
-            onoe <<~EOS
-              No known CI provider detected! If you are using GitHub Actions then we cannot find the expected  environment variables! Check you have e.g. exported them to a Docker container.
-            EOS
-          end
-          diff_end_sha1 = diff_start_sha1 = current_sha1
+        elsif ENV["CI"]
```
MikeMcQuaid, line 111: GitHub Actions always sets this but it made not be passed through to e.g. Docker correctly. This is potentially an issue with `GITHUB_REPOSITORY` too so we may want to fail harder if we find _some_ GitHub Actions variables but not others (because the failure case is otherwise a warning and us building the wrong stuff...)

lib/tests/formulae.rb
Bo98: I mean - does it even need to apply when `GITHUB_SHA` isn't present (i.e. not GitHub Actions)? I don't think we need to `merge-base` in that scenario either.

lib/tests/formulae.rb
Bo98: I don't want the `else` clause if the tap doesn't match, so I'll need to duplicate the `ENV["GITHUB_REPOSITORY"].blank?` check in this case.

lib/tests/formulae.rb
Bo98: So `ENV["CI"] || ENV["GITHUB_ACTIONS"]` like before or?

MikeMcQuaid: Yeh, alternatively have a separate `ENV["GITHUB_ACTIONS"]` case which does a `raise` or `odie` because if we can't find the right environment variables: things are 💥

lib/tests/formulae.rb
MikeMcQuaid: Works for me 👍🏻

lib/tests/formulae.rb
```diff
@@ -82,27 +82,25 @@ def detect_formulae!(args:)
...
         end
 
-        # Use GitHub Actions variables for pull request jobs.
-        if ENV["GITHUB_BASE_REF"].present? && ENV["GITHUB_SHA"].present?
-          diff_start_sha1 = rev_parse("origin/#{ENV["GITHUB_BASE_REF"]}")
-          if diff_start_sha1.blank?
-            test git, "-C", repository, "fetch", "--depth=1",
-                 "origin", "+refs/heads/#{ENV["GITHUB_BASE_REF"]}",
-                 args: args
+        if tap.present? && ENV["GITHUB_REPOSITORY"].present? && tap.full_name.casecmp(ENV["GITHUB_REPOSITORY"]).zero?
```
MikeMcQuaid, line 85:
```suggestion
        if tap.present? &&
           ENV["GITHUB_SHA"].present? &&
           ENV["GITHUB_REPOSITORY"].present? &&
           tap.full_name.casecmp(ENV["GITHUB_REPOSITORY"]).zero? &&
```

lib/tests/formulae.rb
```diff
@@ -82,27 +82,25 @@ def detect_formulae!(args:)
...
-          if ENV["CI"] || ENV["GITHUB_ACTIONS"]
-            onoe <<~EOS
-              No known CI provider detected! If you are using GitHub Actions then we cannot find the expected  environment variables! Check you have e.g. exported them to a Docker container.
-            EOS
-          end
-          diff_end_sha1 = diff_start_sha1 = current_sha1
+        elsif ENV["GITHUB_REPOSITORY"].blank? && (ENV["CI"] || ENV["GITHUB_ACTIONS"])
+          onoe <<~EOS
+            No known CI provider detected! If you are using GitHub Actions then we cannot find the expected environment variables! Check you have e.g. exported them to a Docker container.
+          EOS
```
MikeMcQuaid, line 103:
```suggestion
        elsif ENV["CI"]
          onoe <<~EOS
            No known CI provider detected! If you are using GitHub Actions then we cannot find the expected environment variables! Check you have e.g. exported them to a Docker container.
          EOS
        elsif ENV["GITHUB_ACTIONS"]
           odie <<~EOS
            We cannot find the needed GitHub Actions environment variables! Check you have e.g. exported them to a Docker container.
          EOS
        end
```

Bo98: Still need to make sure we don't do these if it's the tap comparison which failed, but I think I can just reorder the checks.

PR #454, opened 2020-08-01
lib/tests/formulae: fix handling of non-core dependencies

Bo98:
`brew deps` is allowed to fail here so we can't use `safe_popen_read`. However, we should still make sure it is failing for reason we think it is and abort otherwise.


MikeMcQuaid:
Nicely comprehensive 👏🏻

PR #455, opened 2020-08-02
Refactor dry run & fail fast handling

Bo98:
I personally found this call very confusing:

```rb
test "command", "arg1", "arg2", args: args
```

I've tried to refactor that away by instead passing the needed dry run and fail fast args to the Test class `initialize` method, store it there and reference that in the `test` method.


Bo98:
I'll fight with RSpec tomorrow. Maybe using `Predicable` from `brew` was a bad idea.

PR #459, opened 2020-08-04
formulae: don't audit deprecated formulae

dtrodrigues:
It might not be worth spending time resolving audit failures of deprecated formulae, so don't have `test-bot` audit them.

Motivated by https://github.com/Homebrew/homebrew-core/pull/59071.


MikeMcQuaid:
Seems reasonable to me 👍🏻 but note this will only affect some audits. `brew audit --skip-style` is run on all formulae in every PR (but these are all passing anyway).

PR #460, opened 2020-08-05
Better support and use Homebrew actions

MikeMcQuaid:
- use Homebrew/actions/setup-homebrew
- do dry-run tests on Linux too
- move some GitHub Actions cleanup logic to cleanup_before
- only run `brew audit` in `tap_syntax` if there's formula files

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,69 +13,45 @@ jobs:
     - name: Set up Git repository
       uses: actions/checkout@master
 
-    - name: Set up Homebrew
-      id: set-up-homebrew
-      run: |
-        if which brew &>/dev/null; then
-          HOMEBREW_PREFIX="$(brew --prefix)"
-          HOMEBREW_REPOSITORY="$HOMEBREW_PREFIX/Homebrew"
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -13,69 +13,42 @@ jobs:
...
       run: brew install-bundler-gems
 
     - name: Run brew style
       run: brew style homebrew/test-bot
 
+    - name: Cache Homebrew/homebrew-test-bot RubyGems
+      id: cache-test-bot
+      uses: actions/cache@master
+      with:
+        path: vendor/ruby
```
Bo98, line 45: Not entirely sure where `bundle` detects to install here, but to check: is the layout that `setup.rb` is at `vendor/bundler/setup.rb`?

I ask because that's not the case with homebrew-formula-analytics, where the correct path there to cache was `vendor/ruby/ruby`.

MikeMcQuaid: In this case it's just `vendor/ruby`.

MikeMcQuaid: (tested them both out locally)

PR #462, opened 2020-08-05
test_cleanup: allow (but don't require) some taps.

MikeMcQuaid:
These are all taps that will do no harm to be tapped (and we want to use them in Homebrew/brew CI).


MikeMcQuaid:
(as-is this is the wrong revision so https://github.com/Homebrew/actions/pull/39 is needed).

PR #472, opened 2020-08-07
formulae: fully fetch origin/master

dawidd6:
Another try to fix modified formulae detection.


MikeMcQuaid:
Let's try it.

Bo98:
This makes sense to me, I think this will fix things.

If it doesn't: we might need to pass `--unshallow`. Git remembers the last depth passed permanently.

PR #473, opened 2020-08-07
cleanup_before: mv to tmp subdirectory instead of just tmp

SeekingMeaning:


Bo98:
This step is only meant to run on environments that are ephemeral and should never be run more than once.

I do agree that putting it into one folder rather than various folders in `/tmp` is a bit safer to do however (regardless if constant or random).

PurpleBooth:
This is probably the appropriate thing to use - https://ruby-doc.org/stdlib-2.5.3/libdoc/tmpdir/rdoc/Dir.html#method-c-mktmpdir

lib/tests/cleanup_before.rb
```diff
@@ -28,7 +28,11 @@ def run!(args:)
           elsif OS.mac?
             if Dir.glob("#{HOMEBREW_CELLAR}/*").present?
               # moving is much faster than deleting.
-              test "bash", "-c", "mv #{HOMEBREW_CELLAR}/* /tmp"
+              tmp_dir = "/tmp/homebrew-cellar"
+              # tmp_dir shouldn't exist but try deleting it just in case
+              test "bash", "-c", "rm -rf #{tmp_dir}"
```
MikeMcQuaid, line 33: This would remove any speed benefits from moving because we're deleting and moving.

MikeMcQuaid:
> This step is only meant to run on environments that are ephemeral and should never be run more than once. It's a massive waste of space to do so on runners which are not ephermeral.

Correct.

> I do agree that putting it into one subfolder of `/tmp` rather than various ones is a bit safer to do however (regardless if constant or random).

`/tmp/test-bot-cleanup` or similar.



> This is probably the appropriate thing to use - https://ruby-doc.org/stdlib-2.5.3/libdoc/tmpdir/rdoc/Dir.html#method-c-mktmpdir

It is not because it will be a different path every time.

lib/tests/cleanup_before.rb
```diff
@@ -28,9 +28,8 @@ def run!(args:)
           elsif OS.mac?
             if Dir.glob("#{HOMEBREW_CELLAR}/*").present?
               # moving is much faster than deleting.
-              tmp_dir = "/tmp/test-bot-cleanup"
-              test "bash", "-c", "mkdir -p #{tmp_dir}"
-              test "bash", "-c", "mv #{HOMEBREW_CELLAR}/* #{tmp_dir}"
+              test "bash", "-c", "mkdir -p /tmp/test-bot-cleanup"
```
MikeMcQuaid, line 31:
```suggestion
              FileUtils.mkdir_p "/tmp/test-bot-cleanup"
```

Untested but not reason to shell out here.

lib/tests/cleanup_before.rb
```diff
@@ -28,9 +28,8 @@ def run!(args:)
           elsif OS.mac?
             if Dir.glob("#{HOMEBREW_CELLAR}/*").present?
               # moving is much faster than deleting.
```
MikeMcQuaid, line 30: This comment should be above the actual move.

lib/tests/cleanup_before.rb
SeekingMeaning: I think that should work

Bo98:
Thanks!

Issue #302, opened 2019-12-02
Can't have other branch than master while uploading


dawidd6:
Using Github Actions, if one runs `brew test-bot --ci-upload`, the tap is automatically resetted to `master`. Which is undesirable in situations when one checkouts the repo (tap) manually to, let's say, another branch, before uploading the bottles. This is of course not a problem for taps, that rely on `test-bot` for pushing tags, but it is broken for the ones that push branches for example.

Interesting lines of code:
https://github.com/Homebrew/homebrew-test-bot/blob/af6fcf3ef7118acd72e2eb94a9b49b45e8433d67/lib/test_bot.rb#L161-L178

What I'm talking about is demonstrated here: https://github.com/dawidd6/homebrew-tap/pull/71/checks?check_run_id=329233117#step:8:1

I didn't encounter this issue before, cause I didn't try to add a new formula and build a bottle for it via PR.

MikeMcQuaid:
> Which is undesirable in situations when one checkouts the repo (tap) manually to, let's say, another branch, before uploading the bottles.

Can you elaborate why you'd want/need to do this?

dawidd6:
The core reason is that if one of those env variables are not set:
```ruby
 pr = ENV["UPSTREAM_PULL_REQUEST"] || 
     ENV["CHANGE_ID"] || 
     ENV["CIRCLE_PR_NUMBER"] 
```

the required `brew pull` step will not execute and if one for example already has the tap checked out to PR head branch, it will still be resetted to `master`:
```
safe_system "git", "checkout", "-f", "master" 
```

of course, one those env variables could be set, but then the `git tag` and `git push` steps will be executed and I don't want that.

Now that I'm thinking about it, the best solution would be to provide a new option: `--no-push`. Using this option, I would be able to set for example `CHANGE_ID` and `test-bot` would then execute `brew pull`, but won't `git tag` and `git push`. Then I could easily push the new `pr-` branch with bottle commit as usual.

MikeMcQuaid:
@dawidd6 How about moving all that logic into `if pr`?

dawidd6:
Yes, that might actually help. Will do the PR.

dawidd6:
Thanks for merging the PR! It works now.

PR #475, opened 2020-08-09
Improve the readme

PurpleBooth:
While this tool is used by core, the contributors who setup core and
maintain it are closely linked to homebrew project. As such they have in
depth knowledge of the domain and so are less likely to need this
document. There's also only one core (compared to the large number of
custom taps), as such this documentation may be helpful to core, but
it's not it's primary audience.

This document is written for the audience developers who don't have 100
stars (or whaver the required limit is) to get there formula into core,
and as such are creating a custom tap to distribute their application.

The key questions it aims to answer for those custom tap developers are:
* How do I test my new formula
* How to a automate the tests to that custom formula so people can
  contribute
* How do I build binaries for my formula so installs are fast

These are all tasks that test-bot significantly makes simpler, but
pretty much no one I talk to knows about it because these great features
aren't easily discoverable.

There are also a number of idiosyncrasies that happen on GitHub Actions
that need to be documented for a smooth "oh wow it just works"
experience for first time users.

This is also our opportunity to state to the user what we expect you to
do, and, by ommission, what we don't expect, hopefully reducing
questions like #474, #302, and #251

This should also live in the main man pages for the `test-bot` tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,97 @@
-# Homebrew Test Bot
+Homebrew Test Bot
+=================
 
-Tests the full lifecycle of a Homebrew change to a tap (Git repository).
+Homebrew comes with a comprehensive suite of tools that lets you
+validate that both your homebrew installation is working and prove that
+a formula is working.
 
-## Install
+Homebrew Test Bot uses these tools to simplify the automation of your
+formula publishing process including Checking your installation,
+Checking your Formula, and Building and uploading binaries.
+
+Install
+-------
 
 `brew test-bot` is automatically installed when first run.
 
-## Usage
+Usage
+-----
 
-See [the `brew test-bot` section of the `brew man` output](https://docs.brew.sh/Manpage#test-bot-options-formula) or `brew test-bot --help`.
+Here are some of the most common use-cases for `brew test-bot`
 
-## Tests
+### Locally Validating a formula
 
-Tests can be run with `bundle install && bundle exec rspec`.
+If this is for a custom tap, first, tap the repository.
 
-## Copyright
+    brew tap "custom/tap"
 
-Copyright (c) Homebrew maintainers. See [LICENSE.txt](https://github.com/Homebrew/homebrew-test-bot/blob/HEAD/LICENSE.txt) for details.
+Then make your changes.
+
+    echo $YOUR_CHANGES > "$(brew --repo)/Library/Taps/custom/homebrew-tap/Formula/your-formula.rb"
+
+Next, we can validate our changes, including testing and linting, and
+dependency checks.
+
+    brew test-bot "custom/tap/your-formula"
+
+### Validating a formula change in GitHub actions
+
+You are running on GitHub actions test-bot will try to fix any potential
+problems in the homebrew repository deleting the `$(brew --repo)/Cellar`
+before performing any operations. If you've intentionally made changes
+there, they will be lost, what's more `--cleanup` is always passed.
+
+When `--cleanup` is passed, we hard reset the taps to "origin/master",
+as such your changes will need to be cloneable from "master".
+
+If you're using this tool for a personal tap, and as such won't have the
+changes on `master` because you're validating a pull request, or want to
+use another default branch, As a workaround, you can unset the
+`GITHUB_ACTIONS` environment variable to stop this behaviour, and
+manually triggering the clean-up steps.
+
+    brew update-reset
+    brew test-bot --cleanup --only-cleanup-before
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,19 +1,97 @@
-# Homebrew Test Bot
+Homebrew Test Bot
+=================
```
MikeMcQuaid, line 2: Please don't make changes like this, it makes the overall changes hard to follow based on your syntax preferences.

README.md
```diff
@@ -1,19 +1,97 @@
...
 
-Tests the full lifecycle of a Homebrew change to a tap (Git repository).
+Homebrew comes with a comprehensive suite of tools that lets you
+validate that both your homebrew installation is working and prove that
+a formula is working.
```
MikeMcQuaid, line 6:
```suggestion
Homebrew provides tools to verify your Homebrew installation and validate formula installation. Homebrew Test Bot uses these to automate the formula build and test process and upload the produced binaries.
```

README.md
```diff
@@ -1,19 +1,97 @@
...
 
-## Install
+Homebrew Test Bot uses these tools to simplify the automation of your
+formula publishing process including Checking your installation,
+Checking your Formula, and Building and uploading binaries.
```
MikeMcQuaid, line 10:
```suggestion
```

README.md
```diff
@@ -1,19 +1,97 @@
...
 
 `brew test-bot` is automatically installed when first run.
 
-## Usage
+Usage
+-----
+
+Here are some of the most common use-cases for `brew test-bot`
+
+### Locally Validating a formula
```
MikeMcQuaid, line 22: I don't really consider this a common use-case.

README.md
```diff
@@ -1,19 +1,97 @@
...
+    brew test-bot "custom/tap/your-formula"
+
+### Validating a formula change in GitHub actions
+
+You are running on GitHub actions test-bot will try to fix any potential
+problems in the homebrew repository deleting the `$(brew --repo)/Cellar`
+before performing any operations. If you've intentionally made changes
+there, they will be lost, what's more `--cleanup` is always passed.
+
+When `--cleanup` is passed, we hard reset the taps to "origin/master",
```
MikeMcQuaid, line 44: I'd rather we point to (and update, if necessary) `--help` instead of documenting behaviour here that may change in future.

README.md
```diff
@@ -1,19 +1,97 @@
...
+as such your changes will need to be cloneable from "master".
+
+If you're using this tool for a personal tap, and as such won't have the
+changes on `master` because you're validating a pull request, or want to
+use another default branch, As a workaround, you can unset the
+`GITHUB_ACTIONS` environment variable to stop this behaviour, and
```
MikeMcQuaid, line 50: I don't want to recommend this here.

README.md
```diff
@@ -1,19 +1,97 @@
...
+Then push the changes it's made to the tap to include the download
+information in the formula.
+
+    cd "$(brew --repo)/Library/Taps/custom/homebrew-tap"
+    git push
+
+### Full options
 
-See [the `brew test-bot` section of the `brew man` output](https://docs.brew.sh/Manpage#test-bot-options-formula) or `brew test-bot --help`.
+See [the `brew test-bot` section of the `brew man`
```
MikeMcQuaid, line 83: Please revert this.

README.md
```diff
@@ -1,19 +1,97 @@
...
+output](https://docs.brew.sh/Manpage#test-bot-options-formula) or
+`brew test-bot --help`.
 
-## Tests
+Tests
+-----
 
-Tests can be run with `bundle install && bundle exec rspec`.
+You can run the tests with `bundle install && bundle exec rspec`.
```
MikeMcQuaid, line 90: Please revert this.

MikeMcQuaid:
> This should also live in the main man pages for the `test-bot` tool.

If so: they should not be duplicated here and should instead be in https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/test-bot.rb.



> This document is written for the audience developers who don't have 100
> stars (or whaver the required limit is) to get there formula into core,
> and as such are creating a custom tap to distribute their application.

They should be using `brew tap-new` to setup their custom tap which will configure `brew test-bot` for them (and remove most of the need for the manual steps here). Can we point to that instead?

README.md
PurpleBooth: I think I'll swap out this whole section to just a be a reference to the `tap-new` command

PR #476, opened 2020-08-09
Don't assume the default branch is master

PurpleBooth:
This works out what the actual default branch is before resetting

### DIFF
--- a/lib/test_cleanup.rb
+++ b/lib/test_cleanup.rb
@@ -31,9 +31,11 @@ module Homebrew
     ]).freeze
 
     def reset_if_needed(repository)
-      return if system(git, "-C", repository, "diff", "--quiet", "origin/master")
+      default_ref = default_origin_ref(repository)
 
-      test git, "-C", repository, "reset", "--hard", "origin/master"
+      return if system(git, "-C", repository, "diff", "--quiet", default_ref)
+
+      test git, "-C", repository, "reset", "--hard", default_ref
     end
 
     def cleanup_shared
@@ -104,11 +106,21 @@ module Homebrew
 
     private
 
-    def checkout_branch_if_needed(repository, branch = "master")
+    def default_origin_ref(repository)
+      default_branch = Utils.popen_read(
+        git, "-C", repository, "symbolic-ref", "refs/remotes/origin/HEAD", "--short"
+      ).strip.presence
+      default_branch ||= "origin/master"
+      default_branch
+    end
+
+    def checkout_branch_if_needed(repository)
+      # We limit this to two parts, because branch names can have slashes in
+      default_branch = default_origin_ref(repository).split("/", 2).last
       current_branch = Utils.safe_popen_read(
-        git, "-C", repository, "symbolic-ref", "HEAD"
+        git, "-C", repository, "symbolic-ref", "HEAD", "--short"
       ).strip
-      return if branch == current_branch
+      return if default_branch == current_branch
 
       test git, "-C", repository, "checkout", "-f", branch
     end


lib/test_cleanup.rb
```diff
@@ -31,9 +31,13 @@ class TestCleanup < Test
     ]).freeze
 
     def reset_if_needed(repository)
-      return if system(git, "-C", repository, "diff", "--quiet", "origin/master")
+      default_ref = Utils.safe_popen_read(
+        git, "-C", repository, "symbolic-ref", "refs/remotes/origin/HEAD", "--short"
+      ).strip
```
MikeMcQuaid, line 36: This should be extracted into a function, use `popen_read`, appended with `.presence` (which will make it return `nil` if not `present?`) and be `||=`d to `origin/master` if it's `nil` (as I've found some local repositories where this will fail).

lib/test_cleanup.rb
```diff
@@ -104,11 +108,15 @@ def cleanup_shared
 
     private
 
-    def checkout_branch_if_needed(repository, branch = "master")
+    def checkout_branch_if_needed(repository)
+      default_branch = Utils.safe_popen_read(
+        git, "-C", repository, "symbolic-ref", "refs/remotes/origin/HEAD", "--short"
+      ).strip.split("/", 2).last
```
MikeMcQuaid, line 114: This should use the extracted function

lib/test_cleanup.rb
```diff
@@ -104,11 +108,15 @@ def cleanup_shared
...
       current_branch = Utils.safe_popen_read(
-        git, "-C", repository, "symbolic-ref", "HEAD"
+        git, "-C", repository, "symbolic-ref", "HEAD", "--short"
```
MikeMcQuaid, line 116: Why the `--short` addition?

PurpleBooth: Without the short this is 

```
$ git symbolic-ref HEAD
refs/heads/master
```

Whereas what we're really looking for is just "master" I can't see how it could work currently, but perhaps that's just something about my machine.

lib/test_cleanup.rb
```diff
@@ -104,11 +106,19 @@ def cleanup_shared
...
-    def checkout_branch_if_needed(repository, branch = "master")
+    def default_origin_ref(repository)
+      default_branch = Utils.popen_read(
+        git, "-C", repository, "symbolic-ref", "refs/remotes/origin/HEAD", "--short"
+      ).presence ||= "origin/master"
+      default_branch.strip
+    end
+
+    def checkout_branch_if_needed(repository)
+      default_branch = default_origin_ref(repository).split("/", 2).last
```
MikeMcQuaid, line 117:
```suggestion
      default_branch = default_origin_ref(repository).split("/").last
```

lib/test_cleanup.rb
```diff
@@ -104,11 +106,19 @@ def cleanup_shared
 
     private
 
-    def checkout_branch_if_needed(repository, branch = "master")
+    def default_origin_ref(repository)
+      default_branch = Utils.popen_read(
+        git, "-C", repository, "symbolic-ref", "refs/remotes/origin/HEAD", "--short"
+      ).presence ||= "origin/master"
+      default_branch.strip
```
MikeMcQuaid, line 113:
```suggestion
      default_branch = Utils.popen_read(
        git, "-C", repository, "symbolic-ref", "refs/remotes/origin/HEAD", "--short"
      ).strip.presence
      default_branch ||= "origin/master"
      default_branch
```

lib/test_cleanup.rb
PurpleBooth: I did it this way because you can have "/" in branch names, which is confusing but it's the way it is

lib/test_cleanup.rb
PurpleBooth: Won't strip fail if the command fails? (My ruby knowledge is failing me)

lib/test_cleanup.rb
MikeMcQuaid: Cool. Can you leave a comment explaining this?

lib/test_cleanup.rb
MikeMcQuaid: No, it'll always return a string.

lib/test_cleanup.rb
PurpleBooth: Sure!

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @PurpleBooth!

PR #477, opened 2020-08-10
formulae: use full names

SeekingMeaning:
This updates some commands that aren't already using full names, which fixes an issue with bottling of having a formula that has the same name as a core formula


MikeMcQuaid:
Thanks @SeekingMeaning!

SeekingMeaning:
Ouch this broke stuff

MikeMcQuaid:
Reverting on https://github.com/Homebrew/homebrew-test-bot/pull/479

PR #480, opened 2020-08-10
formulae: use full names

SeekingMeaning:
Redo of #477, which will be reverted by #479 as `Dependency` and `FormulaConflict` do not have a `full_name` method (`Dependency#name` and `FormulaConflict#name` already return the full name)


SeekingMeaning:
Is this good to merge?

MikeMcQuaid (approved):
👍🏻 if you're confident.

SeekingMeaning:
`formula.full_name` is used a few lines down so I'm 99.9% confident

PR #486, opened 2020-08-22
formulae: set root_url to GitHub Releases url

dawidd6:
If not testing core tap (and a bunch of other conditions), set the root_url to GitHub releases url.
This url would later be used in `brew pr-upload` to determine what release to create.

### DIFF
--- a/lib/test_ci_upload.rb
+++ b/lib/test_ci_upload.rb
@@ -37,6 +37,14 @@ module Homebrew
         hash.deep_merge(JSON.parse(IO.read(json_file)))
       end
 
+      bottles_hash.each do |_, bottle_hash|
+        root_url = bottle_hash["bottle"]["root_url"]
+
+        next unless root_url.match HOMEBREW_RELEASES_URL_REGEX
+
+        odie "Refusing to upload to GitHub Releases, use `brew pr-upload`."
+      end
+
       if args.dry_run?
         bottles_hash = {
           "testbottest" => {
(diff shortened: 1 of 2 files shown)


MikeMcQuaid (approved):
Makes sense to me (minus @Bo98's question I'd like to see figured out). `--ci-upload` should probably be configured to fail with a helpful message given this case, too?

dawidd6:
Yes, `--keep-old` was a bad idea, deleted it.

Also added a check for `--ci-upload` to refuse uploading to GitHub Releases.

lib/test_ci_upload.rb
```diff
@@ -37,6 +37,16 @@ def run!(tap, args:)
         hash.deep_merge(JSON.parse(IO.read(json_file)))
       end
 
+      bottles_hash.each do |_, bottle_hash|
+        root_url = bottle_hash["bottle"]["root_url"]
+        url_match = root_url.match HOMEBREW_RELEASES_URL_REGEX
+        _, _, _, tag = *url_match
+
+        next unless tag
```
MikeMcQuaid, line 45:
```suggestion
        next unless root_url.match?(HOMEBREW_RELEASES_URL_REGEX)
```

lib/test_ci_upload.rb
```diff
@@ -37,6 +37,16 @@ def run!(tap, args:)
...
+
+        odie "Cowardly refusing to upload to GitHub Releases, use `brew pr-upload`."
```
MikeMcQuaid, line 47:
```suggestion
        odie "Refusing to upload to GitHub Releases, use `brew pr-upload`."
```

MikeMcQuaid (approved):

PR #491, opened 2020-09-01
formulae: use allowlist terminology

dtrodrigues:
Change from whitelist to allowlist.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -299,7 +299,7 @@ module Homebrew
 
         info_header "Determining dependents..."
 
-        build_dependents_from_source_whitelist = %w[
+        build_dependents_from_source_allowlist = %w[
           cabal-install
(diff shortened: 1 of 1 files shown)


jonchang:
It's not super clear to me what the list is "allowing". Should it be called `build_dependents_from_source` (or something else?)

dtrodrigues:
Yeah it wasn't super clear to me until I read some of the PR history. I was trying to track down something else about how/what test-bot chooses to test when I saw `whitelist` in this file. However, it used to be called `build_dependents_from_source` until https://github.com/Homebrew/homebrew-test-bot/pull/340#discussion_r397266099.

@Bo98 @MikeMcQuaid any suggestions for a new name?

MikeMcQuaid:
No ideas on a better name. Let's go for this for now and can have another PR to improve further. Thanks @dtrodrigues!

PR #496, opened 2020-09-10
formulae: run brew livecheck

dawidd6:
This PR adds a `brew livecheck` test, so we can test it on CI.

cc @samford @nandahkrishna

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -572,6 +572,10 @@ module Homebrew
         fetch_args << "--build-bottle" unless formula.bottle_disabled?
         fetch_args << "--force" if args.cleanup?
 
+        livecheck_args = [formula_name]
+        livecheck_args << "--full-name"
+        livecheck_args << "--debug"
+
         new_formula = @added_formulae.include?(formula_name)
         audit_args = [formula_name, "--online"]
         if new_formula
@@ -611,6 +615,8 @@ module Homebrew
         test "brew", "install", *install_args,
              env:  { "HOMEBREW_DEVELOPER" => nil }
         install_passed = steps.last.passed?
+
+        test "brew", "livecheck", *livecheck_args unless formula.livecheck.skip?
 
         test "brew", "audit", *audit_args unless formula.deprecated?
         return unless install_passed


MikeMcQuaid:
Makes sense, thanks @dawidd6!

samford:
@dawidd6 @MikeMcQuaid I'm just now getting to this but what do we think about adding a `CI-skip-livecheck` label that would skip the `brew livecheck` test on CI?

We still have around 600 formulae that need a `livecheck` block for the check to work properly, so these would seemingly always fail CI due to this test.

However, my bigger concern is PRs that modify more than one formula. If a PR modifies a lot of formulae, then the `brew livecheck` test may take some time and may not be worth it in the context of the PR. If the PR doesn't make any livecheck-related changes, it may make sense to simply skip the `brew livecheck` tests. We had issues with CI running into GitHub's rate limiting (for GitHub-related checks) with a smaller number of PRs, so I imagine it will be worse in homebrew-core.

Similarly, if a PR modifies a lot of formulae and we know it's going to go through a number of iterations before being merged, then we may want to use the `CI-skip-livecheck` label to avoid running the `brew livecheck` test across all the formulae every time the branch is updated.

In the future, it may be better to only run the `brew livecheck` test when a `livecheck` block is added or modified in a PR but that may be challenging to implement.

---

Besides that, in the homebrew-livecheck repository, we only ran `brew livecheck` once on CI, using all the modified formulae as the argument. With the setup in this PR, it seems like `brew livecheck` would be run individually for each formula in the PR. From some very simple testing, running `brew livecheck` once per formula seems to be at least 50% slower (due to the startup time), which would add up for PRs that modify a lot of formulae. It's not a pressing issue but it may be worth revisiting in the future, if it's feasible.

MikeMcQuaid:
> We still have around 600 formulae that need a `livecheck` block for the check to work properly, so these would seemingly always fail CI due to this test.

As-of https://github.com/Homebrew/homebrew-test-bot/pull/497 they won't be run unless they have a block.



> However, my bigger concern is PRs that modify more than one formula. If a PR modifies a lot of formulae, then the `brew livecheck` test may take some time and may not be worth it in the context of the PR. If the PR doesn't make any livecheck-related changes, it may make sense to simply skip the `brew livecheck` tests. We had issues with CI running into GitHub's rate limiting (for GitHub-related checks) with a smaller number of PRs, so I imagine it will be worse in homebrew-core.

Let's hold off on solving this until we see it a problem.



> It's not a pressing issue but it may be worth revisiting in the future, if it's feasible.

This applies similarly to e.g. `brew audit` and friends. I'm not sure it can be easily rectified without larger changes to the `test-bot` code.

PR #501, opened 2020-09-16
Fix name of tap argument

scpeters:
Tap specified in `--tap=` was ignored due to typo.


reitermarkus (approved):

MikeMcQuaid:
Oops, good catch, thanks @scpeters!

scpeters:
one of the joys of dynamic languages :sweat_smile:

PR #520, opened 2020-11-14
Add --skip-dependents option

fxcoudert:
In previous mass bottling workflows (Jenkins 👿 ), we would not systematically test dependents of a formula being bottled.

Doing so takes a lot of our time and CPU power, and some formulas would take days to test all dependents. Moreover, we do not gain much by this testing: most of those failures will be preexisting failures (which we will flag later on during bottling, anyway), or stuff where the Catalina bottle doesn’t work well on Big Sur (which we can’t actually fix anyway until we have bottled dependencies).

This adds a `--skip-dependents` option to restore this previous workflow.

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -49,6 +49,8 @@ module Homebrew
              description: "publish the uploaded bottles."
       switch "--skip-recursive-dependents",
              description: "only test the direct dependents."
+      switch "--skip-dependents",
+             description: "do not test dependents."
       switch "--only-cleanup-before",
              description: "Only run the pre-cleanup step. Needs `--cleanup`."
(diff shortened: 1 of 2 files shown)


fxcoudert:
Error on Linux. Cannot reproduce locally, do not know if it would fail on macOS because it aborts:

```
Error: Failure while executing; `brew uses --include-build --include-test --recursive testbottest` exited with 1. Here's the output:
Error: Process completed with exit code 1.```

dawidd6:
Yes. This error is recently more frequent on Linux. It happens on non-core taps. Dunno what is causing that tho.

fxcoudert:
It appears the test-bot already detects the case I'm trying to deal with. Closing.

PR #527, opened 2020-11-28
Move files where possible instead of deleting them.

reitermarkus:
Add helper function for (re)moving files depending on whether the runner is ephemeral or not.

### DIFF
--- a/lib/test_cleanup.rb
+++ b/lib/test_cleanup.rb
@@ -39,6 +39,24 @@ module Homebrew
       test git, "-C", repository, "reset", "--hard", default_ref
     end
 
+    # Moving files is faster than removing them,
+    # so move them if the current runner is ephemeral.
+    def delete_or_move(paths)
+      paths = Array(paths)
+
+      return if paths.empty?
+
+      symlinks, paths = paths.partition(&:symlink?)
+
+      FileUtils.rm_f symlinks
+
+      if ENV["HOMEBREW_GITHUB_ACTIONS"] && !ENV["GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED"]
+        FileUtils.mv paths, Dir.mktmpdir, force: true
+      else
+        FileUtils.rm_rf paths
+      end
+    end
+
     def cleanup_shared
       cleanup_git_meta(repository)
       clean_if_needed(repository)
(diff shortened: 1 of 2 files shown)


lib/test_cleanup.rb
```diff
@@ -38,6 +38,24 @@ def reset_if_needed(repository)
       test git, "-C", repository, "reset", "--hard", default_ref
     end
 
+    # Moving files is faster than removing them,
+    # so move them if the current runner is ephemeral.
+    def delete_or_move(paths)
+      paths = Array(paths)
+
+      return if paths.empty?
```
MikeMcQuaid, line 46:
```suggestion
    def delete_or_move(paths)
      return if paths.blank?
```

lib/tests/cleanup_before.rb
```diff
@@ -25,11 +25,7 @@ def run!(args:)
             # brew doctor complains
             test "sudo", "rm", "-rf", "/usr/local/include/node/"
           elsif OS.mac?
-            if Dir.glob("#{HOMEBREW_CELLAR}/*").present?
-              FileUtils.mkdir_p "/tmp/test-bot-cleanup"
-              # moving is much faster than deleting.
-              test "bash", "-c", "mv #{HOMEBREW_CELLAR}/* /tmp/test-bot-cleanup"
-            end
+            delete_or_move Pathname.glob(HOMEBREW_CELLAR/"*")
```
MikeMcQuaid, line 28: Could remove the `if ENV["HOMEBREW_GITHUB_ACTIONS"] && !ENV["GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED"]` check on line 22 given `delete_or_move` already handles this?

MikeMcQuaid (commented):
Nice 👏🏻

lib/tests/cleanup_before.rb
reitermarkus: No, because the Linux case doesn't use `delete_or_move`.

MikeMcQuaid: @reitermarkus Might be worth testing if it could?

reitermarkus: Linux needs `sudo`.

fxcoudert:
Hi @reitermarkus I'm trying to figure out what caused this issue: https://github.com/Homebrew/homebrew-core/issues/66328
(dispatch-build-bottle workflow does not clean the node its runs on). There have been relatively few changes to the testbot recently, so could you please have a look at the issue and tell me if you think this could be the cause? Thanks

PR #532, opened 2020-12-06
Exclude casks when determining dependants

Rylan12:
Only include formula dependents when determining which formulae to test. This solves the problems we've seen in https://github.com/Homebrew/homebrew-core/pull/66268 and https://github.com/Homebrew/homebrew-core/pull/66287.

This can't be merged until https://github.com/Homebrew/brew/pull/9437 is merged (which is what adds the `--formula` and `--cask` args to `brew uses`

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -312,10 +312,10 @@ module Homebrew
           rust
         ]
 
-        uses_args = []
+        uses_args = %w[--formula --include-build --include-test]
         uses_args << "--recursive" unless args.skip_recursive_dependents?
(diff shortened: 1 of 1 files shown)


Rylan12:
Okay, this is no longer needed to fix CI, but I still think it's a good idea (assuming that https://github.com/Homebrew/brew/pull/9437 is merged). It may help to stop this issue from re-occurring in the future.

MikeMcQuaid (approved):

MikeMcQuaid:
✅ once/if https://github.com/Homebrew/brew/pull/9437 is merged.

PR #561, opened 2021-01-23
Move bottle to subdirectory if linkage or test failed

### LINKED ISSUE #559: Rename bottles instead of deleting them when post-build tests fail
Currently, test-bot deletes bottles that fail the post-build tests (e.g. `brew test` or `brew linkage --test`). This is designed to avoid the shipment of broken bottles. However, it also removes one avenue of investigating these failures: manually downloading the bottle for inspection and local testing.

I propose that test-bot rename the broken bottles instead. I believe the relevant piece of code is

https://github.com/Homebrew/homebrew-test-bot/blob/328a7db428ced7e2971a72d380e5f9ea5883b9ab/lib/tests/formulae.rb#L657-L660

I'll try to get around to this when I have the time for it, but anyone else can feel free to work on this in the meantime.

SeekingMeaning:
This would allow us to create artifacts of failed bottles that can't be accidentally published; i.e. successful bottles would be uploaded to the `bottles` artifact, while failed bottles could be uploaded to a `bottles (failed)` artifact

Closes #559.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -440,16 +440,16 @@ module Homebrew
         @bottle_filename =
           bottle_step.output
                      .gsub(%r{.*(\./\S+#{Utils::Bottles.native_regex}).*}m, '\1')
-        bottle_json_filename =
+        @bottle_json_filename =
           @bottle_filename.gsub(/\.(\d+\.)?tar\.gz$/, ".json")
         bottle_merge_args =
-          ["--merge", "--write", "--no-commit", bottle_json_filename]
+          ["--merge", "--write", "--no-commit", @bottle_json_filename]
         bottle_merge_args << "--keep-old" if args.keep_old? && !new_formula
 
         test "brew", "bottle", *bottle_merge_args
         test "brew", "uninstall", "--force", formula.full_name
 
-        bottle_json = JSON.parse(File.read(bottle_json_filename))
+        bottle_json = JSON.parse(File.read(@bottle_json_filename))
         root_url = bottle_json.dig(formula.full_name, "bottle", "root_url")
         filename = bottle_json.dig(formula.full_name, "bottle", "tags").values.first["filename"]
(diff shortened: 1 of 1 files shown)


lib/tests/formulae.rb
```diff
@@ -655,7 +655,11 @@ def formula!(formula_name, args:)
 
         # Don't test dependents or keep bottle if the formula linkage or test failed.
```
carlocab, line 656:
```suggestion
        # Move bottle and don't test dependents if the formula linkage or test failed.
```

SeekingMeaning: Nice catch!

carlocab:
Closes #559.

SeekingMeaning:
Ninja edit 😉

carlocab:
I tried it as a separate comment... but it didn't get that dashed underline thing under "Closes". So I guess GitHub only recognises it if it's in the first post or in the commit message?

SeekingMeaning:
Yes, I think that's correct

carlocab:
Can we test this with a broken formula to see if it does what's expected?

SeekingMeaning:
Yes

```sh-session
$ brew edit homebrew/test-bot/testbottest # change test block to `system "false"`

$ mkdir test-bot-test-dir

$ cd test-bot-test-dir

$ brew test-bot --only-formulae --test-default-formula
==> Using Homebrew/homebrew-test-bot 027f369 (Move bottle to subdirectory if linkage or test failed)
...

$ ls -R
.:
failed	steps_output.txt

./failed:
testbottest--0.1.mojave.bottle.json  testbottest--0.1.mojave.bottle.tar.gz
```

That was helpful; I found an error and am now pushing a fix

MikeMcQuaid (approved):

MikeMcQuaid:
Makes sense, thanks @SeekingMeaning!

PR #563, opened 2021-01-25
tests/formulae: don't skip new formulae

### LINKED ISSUE #562: test-bot shouldn't skip new formulae on ARM CI
This follows from https://github.com/Homebrew/homebrew-core/pull/69708.

Currently, test-bot does not attempt to build or test formulae using (I think) the following logic:

https://github.com/Homebrew/homebrew-test-bot/blob/86c53a3edab345504ebd2eb863ce87419ed837de/lib/tests/formulae.rb#L578-L585

This requires adding the `CI-force-arm` label so that test-bot will attempt an ARM build for new formulae. We shouldn't need to do this, and it would be good if `test-bot` recognises that a formula is new and attempt a build even if the `CI-force-arm` label is not set.

I'm still mulling over the right approach, so suggestions here would be appreciated.

carlocab:
Closes #562.
Closes https://github.com/Homebrew/homebrew-core/pull/69708.

Thanks, @bayandin.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -574,11 +574,13 @@ module Homebrew
           skip formula.name
           return
         end
+        new_formula = @added_formulae.include?(formula_name)
 
         if Hardware::CPU.arm? &&
            ENV["HOMEBREW_REQUIRE_BOTTLED_ARM"] &&
            !formula.bottled? &&
-           !formula.bottle_unneeded?
+           !formula.bottle_unneeded? &&
+           !new_formula
           opoo "#{formula.full_name} has not yet been bottled on ARM!"
           skip formula.name
           return
@@ -595,7 +597,6 @@ module Homebrew
         livecheck_args << "--full-name"
         livecheck_args << "--debug"
 
-        new_formula = @added_formulae.include?(formula_name)
         audit_args = [formula_name, "--online"]
         if new_formula
           audit_args << "--new-formula"


Rylan12:
What happens if a new formula depends on an existing formula that doesn't have an ARM bottle yet? I would think that we wouldn't want it to happen in this case, right?

carlocab:
It's not a _huge_ deal, as new formulae are not auto-merged anyway. That said, I am open to suggestions for how to implement skipping testing new formulae that depend on unbottled formulae.

Rylan12:
I see what you're saying but it does take extra CI time.

How about we modify the logic here so that a formula is only skipped on ARM if `brew unbottled` shows that there are dependencies that aren't yet bottled on ARM? Then, new formulae can be tested on ARM only if all of it's dependencies are bottled. Additionally, existing formulae are now automatically tested if there's even a possibility that they could work.

How often is it the case where all of a formula's dependencies are all bottled on ARM but the formula still can't build? I'd guess that this is relatively small (or we wouldn't be getting ready to announce official Apple Silicon support) so we can switch to an opt-out approach rather than an opt-in approach. Thoughts?

carlocab:
It does take extra CI time, but the ARM runners are hardly the bottleneck in the testing process. If we wanted to optimise CI, I don't think this is the most productive place to focus time and resources on.

>How often is it the case where all of a formula's dependencies are all bottled on ARM but the formula still can't build?

I'd guess about a third to half of the time I see the `no ARM bottle` label.

>we can switch to an opt-out approach rather than an opt-in approach.

I do like this idea. This is basically what (I think) we're doing anyway: when something looks like it might bottle, we set `CI-force-arm`, and then remove it and re-test later if it fails. Might be better to make use of a `CI-skip-ARM` flag instead.

Rylan12:
> I'd guess about a third to half of the time I see the `no ARM bottle` label.

That's a little higher than I was hoping, but at least it's no more than half 😅

In these cases, though, it sounds like the general process is "`no ARM bottle` --> `CI-force-ARM` --> failure --> remove `CI-force-ARM`". We could save a step by simply doing "failure -> `CI-skip-ARM`".

> I do like this idea. This is basically what (I think) we're doing anyway: when something looks like it might bottle, we set `CI-force-arm`, and then remove it and re-test later if it fails. Might be better to make use of a `CI-skip-ARM` flag instead.

Yeah, it sounds like it's just automating what's already being done and is a logical step forward.

> I think we may not even need to adopt special logic for new formulae if we take this approach.

Agreed

MikeMcQuaid:
> How about we modify the logic here so that a formula is only skipped on ARM if `brew unbottled` shows that there are dependencies that aren't yet bottled on ARM? Then, new formulae can be tested on ARM only if all of it's dependencies are bottled. Additionally, existing formulae are now automatically tested if there's even a possibility that they could work.

Instead of using `brew unbottled` I'd suggest we query the dependencies directly (as we do in a few other places).

Otherwise 👍🏻 from me.

MikeMcQuaid (approved):

MikeMcQuaid:
Merging as-is as this is an improvement on the status quo but agreed we should consider the case @Rylan12 mentioned if we see it pop up more than e.g. once.

carlocab:
Works as expected: https://github.com/Homebrew/homebrew-core/pull/69787

PR #568, opened 2021-02-02
cleanup: speedup cleanup on GitHub Actions.

MikeMcQuaid:
- move instead of deleting taps and node includes (when sensible)
- don't run `brew cleanup` unless self-hosted

### DIFF
--- a/lib/test_cleanup.rb
+++ b/lib/test_cleanup.rb
@@ -103,12 +103,13 @@ module Homebrew
 
       # Keep all "brew" invocations after HOMEBREW_REPOSITORY operations
       # (which cleans up Homebrew/brew)
-      Tap.names.each do |tap_name|
-        next if tap_name == tap&.name
-        next if ALLOWED_TAPS.include?(tap_name)
+      taps_to_remove = Tap.map do |t|
+        next if t.name == tap&.name
+        next if ALLOWED_TAPS.include?(t.name)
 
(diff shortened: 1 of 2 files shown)


lib/test_cleanup.rb
```diff
@@ -103,12 +103,13 @@ def cleanup_shared
 
       # Keep all "brew" invocations after HOMEBREW_REPOSITORY operations
       # (which cleans up Homebrew/brew)
-      Tap.names.each do |tap_name|
+      taps_to_remove = Tap.names.map do |tap_name|
         next if tap_name == tap&.name
```
dawidd6, line 107: Maybe `.select` would fit here better?

lib/test_cleanup.rb
```diff
@@ -103,12 +103,13 @@ def cleanup_shared
 
       # Keep all "brew" invocations after HOMEBREW_REPOSITORY operations
       # (which cleans up Homebrew/brew)
-      Tap.names.each do |tap_name|
+      taps_to_remove = Tap.names.map do |tap_name|
```
dawidd6, line 106: Maybe `.select` would fit here better?

MikeMcQuaid: It's mapping to the `path` value. It could be a `select` then a `map` but this feels a bit neater.

PR #525, opened 2020-11-26
brew bottle: Set HOMEBREW_NO_PATCHELF_RB_WRITE

sjackman:
Use Ruby patchelf to pour bottles, but not to build bottles. Set `HOMEBREW_NO_PATCHELF_RB_WRITE` unless `HOMEBREW_PATCHELF_RB_WRITE` is set.

PR #569, opened 2021-02-02
brew bottle: remove HOMEBREW_NO_PATCHELF_RB_WRITE

danielnachun:
This should revert this commit: https://github.com/Homebrew/homebrew-test-bot/pull/525#issuecomment-769695114 which disabled our Ruby implementation of `patchelf` when building in CI.  Now that we are confident this is working correctly correctly, we can allow the Ruby `patchelf` to be used in CI as well.


MikeMcQuaid:
Thanks @danielnachun!

sjackman:
cc @iMichka `brew bottle` is now using `patchelf.rb`. Heads up in case you see any weirdness.

PR #570, opened 2021-02-03
formulae: create failed dir only if nonexistent

carlocab:
Fixes https://github.com/Homebrew/brew/issues/10498.

I opted for a `File.directory?` check instead of `mkdir_p` since it seemed more transparent.


MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @carlocab!

PR #575, opened 2021-02-17
lib/test_bot: use Bootsnap.

MikeMcQuaid:
This produces massive initialisation speedups when repeatedly running `brew` commands like `brew test-bot` does.

### DIFF
--- a/lib/test_bot.rb
+++ b/lib/test_bot.rb
@@ -52,6 +52,7 @@ module Homebrew
         raise UsageError, "cannot use --cleanup from HOMEBREW_PREFIX as it will delete all output."
       end
 
+      ENV["HOMEBREW_BOOTSNAP"] = "1"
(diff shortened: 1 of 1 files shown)


lib/test_bot.rb
```diff
@@ -52,6 +52,7 @@ def run!(args)
         raise UsageError, "cannot use --cleanup from HOMEBREW_PREFIX as it will delete all output."
       end
 
+      ENV["HOMEBREW_BOOTSNAP"] = "1"
```
carlocab, line 55: Just do double-check, is this no-op on ARM now too? I know it is when you're on portable-ruby, but I wasn't sure about ARM.

MikeMcQuaid: Yes. Homebrew/brew makes it a no-op anywhere it doesn't work.

carlocab (approved):

PR #577, opened 2021-02-24
lib/tests/formulae: fix gcc install infinite loop

Bo98:
`installed_gcc` would always be reset to `false`.


MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @Bo98!

PR #582, opened 2021-03-17
Split formulae step into multiple steps

MikeMcQuaid:
This splits the current `--only-formulae` step into `--only-formulae` and `--only-formulae-detect`. This will allow the failures to be more obvious on whether they are due the detection of the changed formulae.

In a future PR I hope to split out `--only-formulae-bottle` and `--only-formulae-dependents`.  This will allow the failures to be more obvious on whether they are due to the changed formulae failing, the dependents of these formulae (or the detection of the changed formulae).

This future PR work should provide a few benefits:
- in future potentially cache the changed formulae bottles to avoid rebuilding them every time if they were built successfully last time (speeding up large PR rebuilds)
- in future potentially parallelising the dependent testing across multiple jobs/runners
- generally split up a large, hugely complex class

`--only-formulae` should do the same thing it has always done after this PR is merged.

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -59,10 +59,13 @@ module Homebrew
              description: "Only run the tap syntax check step."
       switch "--only-formulae",
              description: "Only run the formulae steps."
+      switch "--only-formulae-detect",
+             description: "Only run the formulae detection steps."
       switch "--only-cleanup-after",
              description: "Only run the post-cleanup step. Needs `--cleanup`."
       conflicts "--only-cleanup-before", "--only-setup", "--only-tap-syntax",
-                "--only-formulae", "--only-cleanup-after"
+                "--only-formulae", "--only-formulae-detect",
+                "--only-cleanup-after"
     end
   end
 
--- a/lib/test_bot.rb
+++ b/lib/test_bot.rb
@@ -98,8 +98,7 @@ module Homebrew
         GIT, "-C", HOMEBREW_REPOSITORY.to_s,
         "log", "-1", "--format=%s"
       ).strip
-      verb = tap ? "Using" : "Testing"
-      puts Formatter.headline("#{verb} Homebrew/brew #{brew_version} (#{brew_commit_subject})", color: :cyan)
+      puts Formatter.headline("Using Homebrew/brew #{brew_version} (#{brew_commit_subject})", color: :cyan)
 
       if tap.to_s != CoreTap.instance.name
         core_revision = Utils.safe_popen_read(
@@ -110,21 +109,17 @@ module Homebrew
       end
 
       if tap
+        tap_github = " #{ENV["GITHUB_REPOSITORY"]}"
         tap_revision = Utils.safe_popen_read(
           GIT, "-C", tap.path.to_s,
           "log", "-1", "--format=%h (%s)"
         ).strip
-        puts Formatter.headline("Testing #{tap.full_name} #{tap_revision}:", color: :cyan)
+        puts Formatter.headline("Testing #{tap.full_name}#{tap_github} #{tap_revision}:", color: :cyan)
       end
 
       ENV["HOMEBREW_GIT_NAME"] = args.git_name || "BrewTestBot"
       ENV["HOMEBREW_GIT_EMAIL"] = args.git_email ||
                                   "1589480+BrewTestBot@users.noreply.github.com"
-
-      puts Formatter.headline("Tap configuration:", color: :cyan)
-      puts "GITHUB_REPOSITORY: #{ENV["GITHUB_REPOSITORY"]}"
-      puts "Core Tap: #{CoreTap.instance}"
-      puts "Found tap: #{tap}"
 
       Homebrew.failed = !TestRunner.run!(tap, git: GIT, args: args)
     ensure
--- a/lib/test_runner.rb
+++ b/lib/test_runner.rb
@@ -5,6 +5,7 @@ require_relative "test"
 require_relative "test_cleanup"
 require_relative "tests/cleanup_after"
 require_relative "tests/cleanup_before"
+require_relative "tests/formulae_detect"
 require_relative "tests/formulae"
 require_relative "tests/setup"
 require_relative "tests/tap_syntax"
@@ -78,6 +79,7 @@ module Homebrew
                  args.only_setup? ||
                  args.only_tap_syntax? ||
                  args.only_formulae? ||
+                 args.only_formulae_detect? ||
                  args.only_cleanup_after?
       !any_only
     end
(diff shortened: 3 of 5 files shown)


lib/test_formulae.rb
```diff
@@ -0,0 +1,138 @@
...
+        end
+
+        changed_formulae = changed_formulae_dependents.sort do |a1, a2|
+          a2[1].to_i <=> a1[1].to_i
+        end
+        changed_formulae.map!(&:first)
+        unchanged_formulae = Array(@formulae) - changed_formulae
+        changed_formulae + unchanged_formulae
+
+        Array(formulae).each(&block)
```
Bo98, line 134:
```suggestion
        (changed_formulae + unchanged_formulae).each(&block)
```

lib/tests/formulae_detect.rb
```diff
@@ -0,0 +1,124 @@
...
+
+        puts <<-EOS
+    url             #{url.presence || "(undefined)"}
+    #{origin_ref}   #{tap_origin_ref_revision.presence || "(undefined)"}
+    HEAD            #{tap_revision.presence || "(undefined)"}
+    diff_start_sha1 #{diff_start_sha1.presence || "(undefined)"}
+    diff_end_sha1   #{diff_end_sha1.presence || "(undefined)"}
+        EOS
+
+        modified_formulae = []
```
Bo98, line 90:
```suggestion
        @modified_formulae = []
```

or adjust the instances below to be without the `@` (this is only used locally and is not an attr).

lib/test_runner.rb
```diff
@@ -116,7 +141,29 @@ def run_tests(tests, args:)
       begin
         tests[:setup]&.run!(args: args)
         tests[:tap_syntax]&.run!(args: args)
-        tests[:formulae]&.run!(args: args)
+
+        formulae, added_formulae, deleted_formulae, built_formulae, test_default_formula = nil
```
Bo98, line 145: Perhaps initially set these to `[]` instead so we don't need `||= []` everywhere.

`test_default_formula` can be set from `args` right away. I'd possibly even pass it in to the constructor instead of passing it around here.

lib/test_runner.rb
```diff
@@ -116,7 +141,29 @@ def run_tests(tests, args:)
...
+
+          formulae = test.formulae
+          added_formulae = test.added_formulae
+          deleted_formulae = test.deleted_formulae
+          built_formulae = test.built_formulae
+          test_default_formula = test.test_default_formula
+        end
+
+        tests[:formulae_bottle]&.run!(args: args)
+        tests[:formulae_dependents]&.run!(args: args)
```
Bo98, line 166: What are these second runs for?

lib/tests/formulae_bottle.rb
```diff
@@ -0,0 +1,386 @@
...
+          deleted_formula!(f)
+        end
+      end
+
+      def bottle!(formula_name, args:)
+        cleanup_during!(args: args)
+
+        test_header(:FormulaeBottle, method: "bottle!(#{formula_name})")
+
+        built_formulae << formula_name
```
Bo98, line 20: We should probably be consistent in whether we use `@` or not-`@` for accessing the attributes.

lib/tests/formulae_dependents.rb
```diff
@@ -0,0 +1,179 @@
...
+          install_bottled_dependent(dependent, args: args) if bottled
+        end
+
+        @bottled_dependents.each do |dependent|
+          install_bottled_dependent(dependent, args: args)
+        end
+      ensure
+        cleanup_bottle_etc_var(formula) if args.cleanup?
+
+        test "brew", "uninstall", "--force", *@unchanged_dependencies if @unchanged_dependencies.present?
```
Bo98, line 64: `cleanup_bottle_etc_var` definitely needs to be in `FormulaeBottle`. Doesn't matter much for dependents.

The uninstall step should also be there too. It does mean having to reinstall the dependencies when testing dependents, which is a little bit of a slowdown, but the uninstall is important to prevent polluting other bottle builds.

lib/tests/formulae_bottle.rb
```diff
@@ -0,0 +1,386 @@
...
+        end
+
+        dependents = dependents.transpose.first.to_a
+        @source_dependents = @source_dependents.transpose.first.to_a
+
+        @testable_dependents = @source_dependents.select(&:test_defined?)
+        @bottled_dependents = with_env(HOMEBREW_SKIP_OR_LATER_BOTTLES: "1") do
+          dependents.select(&:bottled?)
+        end
+        @testable_dependents += @bottled_dependents.select(&:test_defined?)
```
Bo98, line 310: Move all this to `FormulaeDependents` (necessary to access the variables anyway).

lib/tests/formulae_dependents.rb
```diff
@@ -0,0 +1,179 @@
...
+        if OS.linux? &&
+           tap.present? &&
+           tap.full_name == "Homebrew/homebrew-core" &&
+           ENV["HOMEBREW_REQUIRE_BOTTLED_LINUX"] &&
+           !formula.bottled? &&
+           !formula.bottle_unneeded?
+          opoo "#{formula.full_name} has not yet been bottled on Linux!"
+          skip formula.name
+          return
+        end
```
Bo98, line 45: Would be nice to share this logic with `FormulaeBottle`.

cmd/test-bot.rb
```diff
@@ -59,6 +59,10 @@ def test_bot_args
              description: "Only run the tap syntax check step."
       switch "--only-formulae",
              description: "Only run the formulae steps."
+      switch "--only-formulae-bottle",
+             description: "Only run the formulae bottle step."
+      switch "--only-formulae-dependents",
+             description: "Only run the formulae dependents step."
```
maxim-belkin, line 65: 1. It seems like these options should conflict with each other.
2. There will be at least three `--only-formulae...` switches. I think it makes sense to convert it to a flag.

MikeMcQuaid:
@Bo98 makes sense. feel free to commit these directly to this branch.

Bo98:
TODO (probably in a separate PR first):

* Make `install_gcc_if_needed` smarter. I made it fail a bit harder (it was previously possible for it to "skip" but actually keep going) in the commit I added, but that might break a couple things if we stick with the assumption that GCC is the solution to everything. Alternative could also be to only check it after installing dependencies.

lib/test_bot.rb
```diff
@@ -110,22 +109,18 @@ def run!(args)
...
       end
 
       ENV["HOMEBREW_GIT_NAME"] = args.git_name || "BrewTestBot"
       ENV["HOMEBREW_GIT_EMAIL"] = args.git_email ||
                                   "1589480+BrewTestBot@users.noreply.github.com"
 
-      puts Formatter.headline("Tap configuration:", color: :cyan)
-      puts "GITHUB_REPOSITORY: #{ENV["GITHUB_REPOSITORY"]}"
-      puts "Core Tap: #{CoreTap.instance}"
-      puts "Found tap: #{tap}"
```
MikeMcQuaid, line 127: @iMichka FYI I've rolled this into the existing information:
- CoreTap was already output unless it was the tap being tested (in which case it says "Testing" rather than "Using")
- I've added `GITHUB_REPOSITORY` output if it's set
- Whether the tap was found is indicated by "Testing" in the tap output

PR #583, opened 2021-03-18
lib/test_ci_upload: fix error after brew changes

Bo98:
Probably worth deprecating this at some point, signposting taps to use `brew pr-upload`.

### DIFF
--- a/lib/test_ci_upload.rb
+++ b/lib/test_ci_upload.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "github_releases"
+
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
@Bo98 How did you notice this was broken? Where do we/anyone else use this at this stage?

MikeMcQuaid:
Ah, I see it's almost any run of `brew test-bot`. Ouch. Opened https://github.com/Homebrew/homebrew-test-bot/pull/584 to deprecate this.

PR #606, opened 2021-05-03
Write `brew bottle` output to text file.

Bo98:
I think this will be useful to not only quickly see the checksums of the bottles, but also see what is preventing the bottle from being relocatable.

Requires https://github.com/Homebrew/brew/pull/11307

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,9 +58,9 @@ jobs:
             docker run \
               -e GITHUB_ACTIONS -e GITHUB_BASE_REF -e GITHUB_REF -e GITHUB_REPOSITORY -e GITHUB_SHA \
               --rm brew \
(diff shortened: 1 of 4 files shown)


MikeMcQuaid (approved):

MikeMcQuaid:
Good idea!

Bo98:
> `cat: bottle_output.txt: No such file or directory`

Works fine locally?

dawidd6:
```
git -C /usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot clean -ff -dx
```

Could this command be responsible?

Bo98:
Possibly yes. I think we should be making this use `--only-formulae`.

MikeMcQuaid (approved):

dawidd6 (approved):

PR #617, opened 2021-06-15
Exclude version audit

samford:
This PR modifies `test-bot` to exclude `ResourceAuditor#audit_version` where `brew audit` is run for existing formulae, so we avoid this audit on core CI. The version audit should still be run for new formulae, as we want to ensure that we catch an unnecessary `version`.

A detailed explanation of the reasoning behind this change can be found in Homebrew/brew#11542. It's fine to merge this test-bot PR independently and the version audit will be skipped after the brew PR is merged (as `ResourceAuditor` will start receiving the `--only` and `--except` options).

I'm not very familiar with `test-bot`'s internals, so any feedback on whether I've modified the correct `brew audit` calls would be appreciated.

### DIFF
--- a/lib/tests/tap_syntax.rb
+++ b/lib/tests/tap_syntax.rb
@@ -12,7 +12,7 @@ module Homebrew
                                 MacOS.active_developer_dir == "/Applications/Xcode.app/Contents/Developer"
         test "brew", "style", tap.name unless broken_xcode_rubygems
 
-        test "brew", "audit", "--tap=#{tap.name}"
+        test "brew", "audit", "--tap=#{tap.name}", "--except=version"
       end
     end
   end


lib/tests/formulae.rb
```diff
@@ -619,7 +619,7 @@ def formula!(formula_name, args:)
         if new_formula
           audit_args << "--new-formula"
         else
-          audit_args << "--git" << "--skip-style"
+          audit_args << "--git" << "--skip-style" << "--except=version"
```
MikeMcQuaid, line 622: I reckon it's maybe worth leaving as-is here.

samford: Removed the change in the latest push 👍

Rylan12 (approved):
Looks good to me!

lib/tests/formulae.rb
samford: @MikeMcQuaid Following up on this now that I've seen this run on homebrew/core CI. Omitting this change is fine for `CI-syntax-only` PRs (which a PR that preemptively adds a redundant `version` should be) but CI will predictably fail on a normal run.

If someone were to submit a PR for a new `ammonite-repl` version between when Homebrew/homebrew-core#78878 is merged and Homebrew/brew#11491 is merged and tagged in a new version, CI would fail with the version audit error in that version bump PR.

If we also want to avoid the version audit error in this scenario, I'll need to create a follow-up PR to add this change.

MikeMcQuaid: > If someone were to submit a PR for a new `ammonite-repl` version between when [Homebrew/homebrew-core#78878](https://github.com/Homebrew/homebrew-core/pull/78878) is merged and [Homebrew/brew#11491](https://github.com/Homebrew/brew/pull/11491) is merged and tagged in a new version, CI would fail with the version audit error in that version bump PR.

That seems reasonable/fine with me 👍🏻

PR #620, opened 2021-06-18
Remove old ENV var to skip unbottled ARM tests

samford:
This is a follow-up to #619, which added support for a `HOMEBREW_SKIP_UNBOTTLED_ARM_TESTS` environment variable, as a way of renaming `HOMEBREW_REQUIRE_BOTTLED_ARM`.

After that PR was merged, we replaced usage of this environment variable in homebrew/core (in the `tests.yml` workflow), which appears to be the only place this is used. With that done, this PR removes support for the old `HOMEBREW_REQUIRE_BOTTLED_ARM` environment variable, as it's no longer used.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -583,10 +583,8 @@ module Homebrew
         end
         new_formula = @added_formulae.include?(formula_name)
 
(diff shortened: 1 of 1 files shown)


Rylan12 (approved):

MikeMcQuaid:
Thanks @samford!

PR #621, opened 2021-06-24
Add JUnit output format

bayandin:
This PR adds (or rather brings back) an ability to save `homebrew-test-bot` test results in JUnit format.
Most of the code is copied from the previous version (from https://github.com/Homebrew/homebrew-test-bot/pull/355) with some small adjustments:
- All JUnit related code moved to `lib/junit.rb`
- JUnit reporter supports filtering tests by name (currently is saves results only for `audit` and `test`)

See https://github.com/Homebrew/brew/pull/11578#pullrequestreview-689543802

### DIFF
--- a/lib/junit.rb
+++ b/lib/junit.rb
@@ -1,0 +1,86 @@
+# frozen_string_literal: true
+
+require "rexml/document"
+require "rexml/xmldecl"
+require "rexml/cdata"
+
+module Homebrew
+  class Junit
+    BYTES_IN_1_MEGABYTE = 1024*1024
+    MAX_STEP_OUTPUT_SIZE = (BYTES_IN_1_MEGABYTE - (200*1024)).freeze # margin of safety
+
+    def initialize(tests)
+      @tests = tests
+    end
+
+    def build(filters: nil)
+      filters ||= []
+
+      @xml_document = REXML::Document.new
+      @xml_document << REXML::XMLDecl.new
+      testsuites = @xml_document.add_element "testsuites"
+
+      @tests.each do |test|
+        testsuite = testsuites.add_element "testsuite"
+        testsuite.add_attribute "name", "brew-test-bot.#{Utils::Bottles.tag}"
+        testsuite.add_attribute "tests", test.steps.count(&:passed?)
+        testsuite.add_attribute "failures", test.steps.count(&:failed?)
+        testsuite.add_attribute "timestamp", test.steps.first.start_time.iso8601
+
+        test.steps.each do |step|
+          next unless filters.any? { |filter| step.command_short.start_with? filter }
+
+          testcase = testsuite.add_element "testcase"
+          testcase.add_attribute "name", step.command_short
+          testcase.add_attribute "status", step.status
+          testcase.add_attribute "time", step.time
+          testcase.add_attribute "timestamp", step.start_time.iso8601
+
+          next unless step.output?
+
+          output = sanitize_output_for_xml(step.output)
+          cdata = REXML::CData.new output
+
+          if step.passed?
+            elem = testcase.add_element "system-out"
+          else
+            elem = testcase.add_element "failure"
+            elem.add_attribute "message",
+                               "#{step.status}: #{step.command.join(" ")}"
+          end
+
+          elem << cdata
+        end
+      end
+    end
+
+    def write(filename)
+      output_path = Pathname(filename)
+      output_path.unlink if output_path.exist?
+      output_path.open("w") do |xml_file|
+        pretty_print_indent = 2
+        @xml_document.write(xml_file, pretty_print_indent)
+      end
+    end
+
+    private
+
+    def sanitize_output_for_xml(output)
+      return output if output.blank?
+
+      # Remove invalid XML CData characters from step output.
+      invalid_xml_pat =
+        /[^\x09\x0A\x0D\x20-\uD7FF\uE000-\uFFFD\u{10000}-\u{10FFFF}]/
+      output.gsub!(invalid_xml_pat, "\uFFFD")
+
+      return output if output.bytesize <= MAX_STEP_OUTPUT_SIZE
+
+      # Truncate to 1MB to avoid hitting CI limits
+      output =
+        truncate_text_to_approximate_size(
+          output, MAX_STEP_OUTPUT_SIZE, front_weight: 0.0
+        )
+      "truncated output to 1MB:\n#{output}"
+    end
+  end
+end
--- a/lib/junit.rb
+++ b/lib/junit.rb
@@ -5,10 +5,9 @@ require "rexml/xmldecl"
 require "rexml/cdata"
 
 module Homebrew
+  # Creates Junit report with only required by BuildPulse attributes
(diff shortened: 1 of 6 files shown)


lib/junit.rb
```diff
@@ -0,0 +1,86 @@
...
+          output = sanitize_output_for_xml(step.output)
+          cdata = REXML::CData.new output
+
+          if step.passed?
+            elem = testcase.add_element "system-out"
+          else
+            elem = testcase.add_element "failure"
+            elem.add_attribute "message",
+                               "#{step.status}: #{step.command.join(" ")}"
+          end
```
MikeMcQuaid, line 50: I think we can probably strip the command output. It was pretty useful for Jenkins but it generates huge XML files that we don't really need/use now.

@jasonrudolph any ideas which of these fields BuildPulse requires/will use/will be useful?

MikeMcQuaid (commented):
Loving this so far! Great work @bayandin!

lib/junit.rb
jasonrudolph: > any ideas which of these fields BuildPulse requires/will use/will be useful?

Let's look at an example XML file from rspec_junit_formatter for reference:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="rspec" tests="2" skipped="0" failures="1" errors="0" time="0.015292" timestamp="2021-06-25T07:48:42-04:00" hostname="j19">
<properties><property name="seed" value="55788"/></properties>
<testcase classname="spec.example_0_spec" name="some-describe some-it" file="./spec/example_0_spec.rb" time="0.014212"><failure message="
expected: false
     got: true

(compared using ==)

Diff:
@@ -1,2 +1,2 @@
-false
+true
" type="RSpec::Expectations::ExpectationNotMetError">Failure/Error: expect(true).to eq false

  expected: false
       got: true

  (compared using ==)

  Diff:
  @@ -1,2 +1,2 @@
  -false
  +true

./spec/example_0_spec.rb:3:in `block (2 levels) in &lt;top (required)&gt;&apos;</failure></testcase>
<testcase classname="spec.example_spec" name="some-other-describe some-other-it" file="./spec/example_spec.rb" time="0.000072"></testcase>
</testsuite>
```

BuildPulse requires:
- [testsuite.name](https://github.com/Homebrew/homebrew-test-bot/blob/db8b7b2c2005438a45d0108d2cc6c6bb62e08057/lib/junit.rb#L25) (which is "rpsec" in the xample above)
- [testcase.name](https://github.com/Homebrew/homebrew-test-bot/blob/db8b7b2c2005438a45d0108d2cc6c6bb62e08057/lib/junit.rb#L34) (which is "some-describe some-it" in the first testcase above)
- [testcase.time](https://github.com/Homebrew/homebrew-test-bot/blob/db8b7b2c2005438a45d0108d2cc6c6bb62e08057/lib/junit.rb#L36) (which is "0.014212" in the first testcase above)
- For a failure, the testcase node needs a child [failure](https://github.com/Homebrew/homebrew-test-bot/blob/db8b7b2c2005438a45d0108d2cc6c6bb62e08057/lib/junit.rb#L47) node (shown in the first testcase above)
- For an error, the testcase node needs a child `error` node. It looks like the current code is [classifying any non-passing testcase as a failure](https://github.com/Homebrew/homebrew-test-bot/blob/db8b7b2c2005438a45d0108d2cc6c6bb62e08057/lib/junit.rb#L47), as opposed to distinguishing between errors and failures. That's totally fine from BuildPulse's perspective. 😅
- For failures and errors, BuildPulse accepts either a `message` attribute and/or `CDATA` to [describe the failure/error](https://twitter.com/jasonrudolph/status/1366447143783960579). (In the first testcase example above, the `message` and the CDATA are the same. I'm not yet sure why rspec_junit_formatter duplicates that info.)

If any of the following items are present, BuildPulse will use them:
- testcase.classname (which is "spec.example_0_spec" in the first testcase above)
- testcase.file (which is "./spec/example_0_spec.rb" in the first testcase above) -- This allows the BuildPulse UI/API to provide a link directly to the file that contains the flaky testcase.

BuildPulse ignores child nodes (like [system-out](https://github.com/Homebrew/homebrew-test-bot/blob/db8b7b2c2005438a45d0108d2cc6c6bb62e08057/lib/junit.rb#L45)) for passing tests.

Does that give you the info you're looking for?

bayandin: Thanks for the information @jasonrudolph! I'll double-check that we have all the required fields. 
For some PRs (like `rust` update for example) we could have quite a lot of failures with long `system-out`s. Are there any limits for the resulting file size? We could split a single JUnit file into several (by `testsute.name` for example), does BuildPulse support it?

MikeMcQuaid: > Does that give you the info you're looking for?

Yes, thanks @jasonrudolph!

@bayandin I think we can just populate the stuff that will be used by BuildPulse?

bayandin: > @bayandin I think we can just populate the stuff that will be used by BuildPulse?

Yep, but it still requires error message:

> For failures and errors, BuildPulse accepts either a message attribute and/or CDATA to describe the failure/error. (In the first testcase example above, the message and the CDATA are the same. I'm not yet sure why rspec_junit_formatter duplicates that info.)

Or you suggest using some short cooked up error message instead of command output?

MikeMcQuaid: @bayandin if we're only doing `audit` and `test` for now: I think that output length should always be fine? For `install` (if we add so now or in future) we probably just set a custom string rather than providing the entire install output?

jasonrudolph: > For some PRs (like `rust` update for example) we could have quite a lot of failures with long `system-out`s. Are there any limits for the resulting file size?

Yes. BuildPulse currently rejects any XML files that are larger than 100 megabytes. 

> We could split a single JUnit file into several (by `testsute.name` for example), does BuildPulse support it?

Yes. You can have as many XML files as you like. :-)

bayandin:
I've removed all the extra data (we could easily bring it back when needed from this or from the previous one PR, thanks git!)
And here's an example of junit report for command `brew test-bot --only-formulae --junit hello zzz`:

```xml
<?xml version='1.0'?>
<testsuites>
  <testsuite name='brew-test-bot.big_sur' timestamp='2021-06-29T22:12:46+01:00'>
    <testcase name='audit hello --online --git --skip-style' status='failed' time='23.11467' timestamp='2021-06-29T22:13:47+01:00'>
      <failure message='failed: brew audit hello --online --git --skip-style'/>
    </testcase>
    <testcase name='test hello' status='failed' time='3.29637' timestamp='2021-06-29T22:14:24+01:00'>
      <failure message='failed: brew test --verbose hello'/>
    </testcase>
  </testsuite>
  <testsuite name='brew-test-bot.big_sur' timestamp='2021-06-29T22:14:41+01:00'>
    <testcase name='audit zzz --online --git --skip-style' status='passed' time='98.960434' timestamp='2021-06-29T22:14:51+01:00'/>
    <testcase name='test zzz' status='passed' time='2.836198' timestamp='2021-06-29T22:16:41+01:00'>
    </testcase>
  </testsuite>
</testsuites>
```

MikeMcQuaid (approved):
Looks great, nice work @bayandin!

PR #625, opened 2021-06-30
tests: modify dependents lists only in linuxbrew-core

iMichka:
The lists of dependents are reset to avoid long-running builds in linuxbrew-core.
Run that part of the code only in linuxbrew-core

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -295,7 +295,9 @@ module Homebrew
         @unchanged_build_dependencies = build_dependencies - @formulae
 
         # Test reverse dependencies for linux-only formulae in linuxbrew-core.
-        if args.keep_old? && formula.requirements.exclude?(LinuxRequirement.new)
+        if tap.full_name == "Homebrew/linuxbrew-core" &&
+           args.keep_old? &&
+           formula.requirements.exclude?(LinuxRequirement.new)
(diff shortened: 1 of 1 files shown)


iMichka:
I am unsure if this will fix the issue we are seeing in homebrew-core with the `ncurses` build, but at least we make sure this special workaround runs only on linuxbrew-core.

lib/tests/formulae.rb
```diff
@@ -295,7 +295,9 @@ def setup_formulae_deps_instances(formula, formula_name, args:)
         @unchanged_build_dependencies = build_dependencies - @formulae
 
         # Test reverse dependencies for linux-only formulae in linuxbrew-core.
-        if args.keep_old? && formula.requirements.exclude?(LinuxRequirement.new)
+        if tap.full_name == "Homebrew/linuxbrew-core" &&
```
bayandin, line 298: This doesn't really work:
```
==> Determining dependencies...
Error: undefined method `full_name' for nil:NilClass
```
https://github.com/Homebrew/brew/pull/11628/checks?check_run_id=2954843691#step:4:25

Did you mean something like this:
```suggestion
        if formula.tap.full_name == "Homebrew/linuxbrew-core" &&
```

bayandin: Will be fixed by https://github.com/Homebrew/homebrew-test-bot/pull/627

PR #626, opened 2021-06-30
tests: add some debug output

iMichka:
This can be reverted if we think it is useless,
but I have been asking myself many times why
some dependents are being tested.

This should also help debug an issue we have right now with
homebrew-core.


dawidd6:
We can try it.

dawidd6 (approved):

MikeMcQuaid (approved):
Makes sense to me.

PR #627, opened 2021-06-30
tests: fix for missing tap

iMichka:


iMichka:
@bayandin I tested the initial change locally and had no issue, so it's weird it failed on CI. I saw that we check for the tap's presence somewhere else, as this code might be called multiple times, and sometimes without tap. I hope this fixes the issue you have seen.

bayandin (approved):
Makes sense!

PR #629, opened 2021-07-01
tests: improve :all bottle detection logic

iMichka:
Remove the check on the tap because I have the feeling that it is not always set.
We can skip :all bottles for the time being on any Linux env, as we don't
use :all bottles there.

Change the boolean detection logic, should work better now.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -376,10 +376,7 @@ module Homebrew
 
         @testable_dependents = @source_dependents.select(&:test_defined?)
         @bottled_dependents = with_env(HOMEBREW_SKIP_OR_LATER_BOTTLES: "1") do
-          if OS.linux? &&
-             tap.present? &&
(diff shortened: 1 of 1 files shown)


iMichka:
@dawidd6 this fixed the `typescript` build: https://github.com/Homebrew/homebrew-core/pull/80354/checks?check_run_id=2960911276

The Linux build is now  skipped there.

It did not fix the recursive dependents issue though, so it looks like my logic is still wrong for that ...

MikeMcQuaid:
@iMichka I'd like this to remain tap-specific, please. Otherwise we're making a pretty big behaviour change for any third-party taps using this.

iMichka:
I'll add the tap-specific checks, now that I got everything fixed and it works like I want. It was easier to me to debug stuff with less parameters that could break things. I'll re-enable the taps check one by one and make sure it does not break anything.

MikeMcQuaid:
Thanks @iMichka!

PR #631, opened 2021-07-01
tests: skip all bottles for Linux only in homebrew-core

iMichka:
Now that everything works as expected, I am adding this
check back (I removed it to make sure it did not break anything in a previous PR)


carlocab (approved):

MikeMcQuaid:
Thanks @iMichka! ❤️

PR #636, opened 2021-07-08
Add more arguments

MikeMcQuaid:
Add more switches and flags to customise `brew test-bot` behaviour.

These will be used to move more logic from `brew test-bot` into labels in the Homebrew/homebrew-core workflow and separate formulae detection, build and dependants testing.

While we're here, also cleanup some ugly output that I noticed.

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -22,6 +22,8 @@ module Homebrew
              description: "Don't check if the local system is set up correctly."
       switch "--build-from-source",
              description: "Build from source rather than building bottles."
+      switch "--build-dependents-from-source",
+             description: "Build dependents from source rather than testing bottles."
       switch "--junit",
              description: "generate a JUnit XML test results file."
       switch "--keep-old",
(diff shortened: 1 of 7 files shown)


cmd/test-bot.rb
```diff
@@ -65,9 +69,20 @@ def test_bot_args
              description: "Only run the formulae dependents steps."
       switch "--only-cleanup-after",
              description: "Only run the post-cleanup step. Needs `--cleanup`."
+      flag   "--testing-formulae=",
+             description: "Use these testing formulae rather than running the formulae detection steps."
+      flag   "--added-formulae=",
+             description: "Use these added formulae rather than running the formulae detection steps."
+      flag   "--deleted-formulae=",
+             description: "Use these deleted formulae rather than running the formulae detection steps."
+      flag   "--skipped-or-failed-formulae=",
```
Rylan12, line 78: @MikeMcQuaid (sorry for the late comment) can these use `comma_array` instead of `flag` to avoid the `.split(",")` later on?

PR #643, opened 2021-07-09
formulae_detect: git fetch refs only in non Homebrew repos

dawidd6:
This should be no longer needed for us since we are using setup-homebrew action.


MikeMcQuaid (approved):

MikeMcQuaid:
Makes sense, thanks @dawidd6!

PR #645, opened 2021-07-10
formulae_detect: git fetch origin only if not official tap

dawidd6:
Continuation of: https://github.com/Homebrew/homebrew-test-bot/pull/643


MikeMcQuaid:
Makes sense thanks @dawidd6!

PR #647, opened 2021-07-12
cmd/test-bot: use `comma_array`.

MikeMcQuaid:
As suggested in https://github.com/Homebrew/homebrew-test-bot/pull/636#discussion_r667354914

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -73,15 +73,15 @@ module Homebrew
              description: "Only run the formulae dependents steps."
       switch "--only-cleanup-after",
              description: "Only run the post-cleanup step. Needs `--cleanup`."
-      flag   "--testing-formulae=",
-             description: "Use these testing formulae rather than running the formulae detection steps."
-      flag   "--added-formulae=",
-             description: "Use these added formulae rather than running the formulae detection steps."
-      flag   "--deleted-formulae=",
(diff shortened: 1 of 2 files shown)


cmd/test-bot.rb
```diff
@@ -73,15 +73,15 @@ def test_bot_args
...
       switch "--only-cleanup-after",
              description: "Only run the post-cleanup step. Needs `--cleanup`."
-      flag   "--testing-formulae=",
-             description: "Use these testing formulae rather than running the formulae detection steps."
-      flag   "--added-formulae=",
-             description: "Use these added formulae rather than running the formulae detection steps."
-      flag   "--deleted-formulae=",
-             description: "Use these deleted formulae rather than running the formulae detection steps."
-      flag   "--skipped-or-failed-formulae=",
-             depends_on:  "--only-formulae-dependents",
```
Rylan12, line 83: Is this intentionally being removed? It looks to me like the only time this is used is if the `formulae-dependents` test runs without the `formulae` test running. Which I believe only happens if passing `--only-formulae-dependents`.

Rylan12 (approved):

cmd/test-bot.rb
MikeMcQuaid: It's not supported for `comma_array` currently.

Rylan12: Ah well in that case, 👍 😅

PR #650, opened 2021-07-19
formulae_dependents: install glibc if needed

dawidd6:
See discussion in: https://github.com/Homebrew/homebrew-core/pull/81485


dawidd6:
Let's try this

PR #653, opened 2021-08-18
test-bot: install brewed curl if needed

iMichka:
This is needed when using brewed curl to fetch and/or audit
formulae

See https://github.com/Homebrew/brew/issues/11705

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -37,6 +37,20 @@ module Homebrew
         retry
       end
 
+      def install_curl_if_needed(formula)
+        %w[Stable HEAD].each do |name|
+          spec_name = name.downcase.to_sym
+          next unless (spec = formula.send(spec_name))
+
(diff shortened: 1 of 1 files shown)


iMichka:
> error: could not lock config file .git/config: Permission denied
Error: Command failed with exit 255: git

iMichka:
This is ready for review. @MikeMcQuaid is there a place where I can ask it to uninstall `curl` when the tests are done?

MikeMcQuaid (approved):

iMichka:
I merged this to test the change in homebrew-core. We can't use the new `:using` strategy before a new brew release is made anyway, so we have a little bit of time.

I would still want to explore the possibility to uninstall brewed curl at the end of the test run of each formula.

MikeMcQuaid:
> I would still want to explore the possibility to uninstall brewed curl at the end of the test run of each formula.

@iMichka I think this should happen automatically with the existing code.

PR #665, opened 2021-09-24
lib/tests/formulae*: unset HOMEBREW_GIT_PATH for git dependents

Bo98:
During formula build or test, we need to unset `HOMEBREW_GIT_PATH` if the formula in question has a `depends_on "git"`, otherwise it will continue to prioritise system git in the PATH.

In these scenarios, it is also safe to assume brewed git is functional, otherwise something's already broken and the build/test is going to fail anyway.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -294,10 +294,14 @@ module Homebrew
 
         # Don't care about e.g. bottle failures for dependencies.
         test "brew", "install", "--only-dependencies", *install_args,
-             env:  { "HOMEBREW_DEVELOPER" => nil }
+             env: { "HOMEBREW_DEVELOPER" => nil }
 
+        env = {}
+        env["HOMEBREW_GIT_PATH"] = nil if deps.any? do |d|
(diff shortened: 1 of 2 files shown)


lib/tests/formulae.rb
```diff
@@ -294,10 +294,14 @@ def formula!(formula_name, args:)
 
         # Don't care about e.g. bottle failures for dependencies.
         test "brew", "install", "--only-dependencies", *install_args,
-             env:  { "HOMEBREW_DEVELOPER" => nil }
+             env: { "HOMEBREW_DEVELOPER" => nil }
 
+        env = {}
+        env["HOMEBREW_GIT_PATH"] = nil if deps.any? do |d|
+          d.name == "git" && (!d.test? || d.build?)
+        end
```
MikeMcQuaid, line 302: Could this get pulled into a function that takes `deps` as an argument and returns `env`? It's used in 4 places.

Bo98:
The conditions differ for a couple so it’s not 4 identical blocks but yeah that makes sense.

>

PR #666, opened 2021-09-27
tests/tap_syntax: check both macOS and Linux dep graphs

Bo98:
Discussion in Homebrew/homebrew-core#85798.


MikeMcQuaid (approved):
I'm fine with this as a short-term workaround until the better approach is implemented.

PR #689, opened 2021-11-04
lib/test_bot: dont pass --full flag to brew tap command

bruuuuuuuce:
https://github.com/Homebrew/homebrew-test-bot/issues/688

The command `brew tap` no longer supports the flag `--full`, causes the error 

```
Calling `brew tap --full` is disabled! There is no replacement.
```


carlocab (approved):
Thanks, @bruuuuuuuce.

PR #706, opened 2021-11-15
lib/step: tweak annotation code/output.

MikeMcQuaid:
- Use early `return`s to reduce indentation
- Use `<details>` to optionally display the full output for the failed command in the annotation.


carlocab (approved):
Thanks Mike ❤️

PR #710, opened 2021-11-16
lib/step: improve GitHub Actions annotations.

MikeMcQuaid:
- clarify that these are GitHub Actions annotations in method name
- don't duplicate the output on GitHub Actions
- better truncate the output
- note the OS and architecture in the title

### DIFF
--- a/lib/step.rb
+++ b/lib/step.rb
@@ -69,13 +69,25 @@ module Homebrew
       puts Formatter.headline(Formatter.error("FAILED"), color: :red) unless passed?
     end
 
-    def emit_annotation(type, message, title, file, line)
+    def in_github_actions?
+      ENV["GITHUB_ACTIONS"].present?
+    end
+
+    def puts_github_actions_annotation(type, message, title, file, line)
+      return unless in_github_actions?
+
+      # Temporarily disable annotations on Linux
       # https://github.com/Homebrew/homebrew-test-bot/issues/712
       return if OS.linux?
-      return if ENV["GITHUB_ACTIONS"].blank?
 
       annotation = GitHub::Actions::Annotation.new(type, message, title: title, file: file, line: line)
       puts annotation
+    end
+
+    def puts_in_github_actions_group(title)
+      puts "::group::#{title}" if in_github_actions?
+      yield
+      puts "::endgroup" if in_github_actions?
     end
 
     def output?
@@ -87,6 +99,16 @@ module Homebrew
     # @return [Float] execution time in seconds
     def time
       end_time - start_time
+    end
+
+    def puts_full_output
+      if in_github_actions?
+        puts_in_github_actions_group("Full #{command_short} output") do
+          puts @output
+        end
+      else
+        puts @output
+      end
     end
 
     def run(dry_run: false, fail_fast: false)
@@ -120,9 +142,7 @@ module Homebrew
 
       puts_result
 
-      output = result.merged_output
-
-      unless output.empty?
+      if (output = result.merged_output.presence)
         output.force_encoding(Encoding::UTF_8)
 
         @output = if output.valid_encoding?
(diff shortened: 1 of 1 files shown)


lib/step.rb
```diff
@@ -137,32 +135,55 @@ def run(dry_run: false, fail_fast: false)
...
+              puts @output
+              next
+            end
+
+            annotation_type = failed? ? :error : :warning
+
+            # GitHub Actions has a 64KB maximum for annotiations. That's a bit
+            # too long so instead let's go for a maximum of 256 lines or 24KB.
+            max_length_start = [@output.length - (24 * 1024), 0].max
+            annotation_output = @output.lines.last(100).join("\n")[max_length_start..]
```
carlocab, line 178: Won't `max_length_start` be the starting point relative to when `output` contains all the lines, and not when you start at line `-100`? What you actually want is `@output.lines.last(100).length - (24 * 1024)` I think.

I suggest you take the last x lines of `output` first and then compute `max_length_start`. `last(100)` also doesn't seem to match the comment which says `256` if I've not misunderstood what's happening here.

MikeMcQuaid: Updated!

lib/step.rb
```diff
@@ -137,32 +135,55 @@ def run(dry_run: false, fail_fast: false)
...
+              puts @output
+              next
+            end
+
+            annotation_type = failed? ? :error : :warning
+
+            # GitHub Actions has a 64KB maximum for annotiations. That's a bit
+            # too long so instead let's go for a maximum of 24KB or 256 lines.
+            max_length_start = [@output.length - (24 * 1024), 0].max
+            annotation_output = @output[max_length_start..].lines.last(256).join("\n")
```
carlocab, line 178: I have two concerns with this truncation scheme.

1. This will truncate output in the middle of a line. Not a huge deal, but we can probably fix this.
2. This means you never get to see the untruncated output in the log, which is often of interest (e.g. the output of `./configure`).

As a solution for 2, I suggest we still dump the full `@output` anyway, but hide it under a fold so we only open it up if we need to. https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#grouping-log-lines

carlocab:
You'll have a merge conflict from #716.

lib/step.rb
MikeMcQuaid: @carlocab I'm punting on the middle-of-line truncation (unless we want to go with just line-based truncation which I'd be fine with).

I've added groups to output both types of output. Let me know what you think.

lib/step.rb
```diff
@@ -69,15 +69,27 @@ def puts_result
       puts Formatter.headline(Formatter.error("FAILED"), color: :red) unless passed?
     end
 
-    def emit_annotation(type, message, title, file, line)
+    def in_github_actions?
+      ENV["GITHUB_ACTIONS"].present?
+    end
+
+    def puts_github_actions_annotation(type, message, title, file, line)
+      return if in_github_actions?
```
carlocab, line 77:
```suggestion
      return unless in_github_actions?
```

lib/step.rb
carlocab: I like this, but won't this only show truncated output in the CI log? I think we still want to see the full one too for this case.

MikeMcQuaid: @carlocab Whoops, fixed.

carlocab (approved):

lib/step.rb
```diff
@@ -120,9 +142,7 @@ def run(dry_run: false, fail_fast: false)
 
       puts_result
 
-      output = result.merged_output
-
-      unless output.empty?
+      if (output = result.merged_output.presence)
```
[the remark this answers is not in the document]
Bo98: (Happened before so there should probably be a comment: #361)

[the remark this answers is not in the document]
MikeMcQuaid: Doh. Agreed on comment: https://github.com/Homebrew/homebrew-test-bot/pull/720

PR #715, opened 2021-11-17
formulae_dependents: prune build deps when linking test deps

FnControlOption:
This improves on #711 and #713 by using `Dependency.prune` to skip build dependencies and their recursive dependencies.

Using `Dependency.skip` instead of `next` is technically not needed because the return value of `recursive_dependencies` is not used. However, it makes the intent here clearer: skip the dependency, but continue iterating over its recursive dependencies so they can be linked too.


carlocab (approved):
Thanks, @FnControlOption! Nice work.

PR #746, opened 2021-12-17
formulae_dependents: run postinstall before testing dependents

Bo98:
This fixes an issue where important etc/var files, which may have been nuked during the build stage, are missing.

Detailed at: https://github.com/Homebrew/homebrew-core/pull/85812#issuecomment-996325061

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,9 +48,14 @@ jobs:
       - name: Run Homebrew/homebrew-test-bot RSpec tests
         run: bundle exec rspec
 
-      - name: Build Docker image
+      - name: Build Docker image and start container
         if: matrix.os == 'ubuntu-latest'
-        run: docker build -t brew .
+        run: |
+          docker build -t brew .
+          docker create \
+            -e GITHUB_ACTIONS -e GITHUB_BASE_REF -e GITHUB_REF -e GITHUB_REPOSITORY -e GITHUB_SHA \
+            --name=brewtestbot brew sleep infinity
+          docker start brewtestbot
 
       - run: brew test-bot --only-formulae-detect --test-default-formula
         id: formulae-detect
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -48,9 +48,14 @@ jobs:
       - name: Run Homebrew/homebrew-test-bot RSpec tests
         run: bundle exec rspec
 
-      - name: Build Docker image
+      - name: Build Docker image and start container
```
MikeMcQuaid, line 51: Can you provide more context for all the Docker changes? They seem unrelated but I presume they are not.

lib/tests/formulae_dependents.rb
```diff
@@ -24,6 +24,10 @@ def dependent_formulae!(formula_name, args:)
           return
         end
 
+        # Restore etc/var files that may have been nuked in the build stage.
+        test "brew", "postinstall", formula_name
```
MikeMcQuaid, line 28: Makes sense, thanks!

lib/tests/formulae_dependents.rb
```diff
@@ -24,6 +24,10 @@ def dependent_formulae!(formula_name, args:)
...
+        return if steps.last.failed?
```
MikeMcQuaid, line 29: Should this error or anything?

.github/workflows/tests.yml
Bo98: The `--only-formulae-dependents` test did nothing because we weren't passing `--testing-formulae`.

When I added it, it revealed that because we're creating a new container for each test then it errored. `--only-formulae-dependents` assumes that the formula is already installed (which can be changed but I think makes sense).

MikeMcQuaid: Gotcha, thanks ❤️

lib/tests/formulae_dependents.rb
MikeMcQuaid: Great 👍🏻

PR #786, opened 2022-02-14
test_bot: allow to override HOMEBREW_CURL_PATH

iMichka:
This is necessary for running tests on older Linux systems,
were we provide a newer curl version which is installed in a different path.

### DIFF
--- a/lib/test_bot.rb
+++ b/lib/test_bot.rb
@@ -57,7 +57,7 @@ module Homebrew
       ENV["HOMEBREW_NO_EMOJI"] = "1"
       ENV["HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK"] = "1"
       ENV["HOMEBREW_FAIL_LOG_LINES"] = "150"
-      ENV["HOMEBREW_CURL_PATH"] = "/usr/bin/curl"
+      ENV["HOMEBREW_CURL_PATH"] = ENV["HOMEBREW_ON_DEBIAN7"].present? ? "/opt/bin/curl" : "/usr/bin/curl"
       ENV["HOMEBREW_GIT_PATH"] = GIT
       ENV["HOMEBREW_PATH"] = ENV["PATH"] =
         "#{HOMEBREW_PREFIX}/bin:#{HOMEBREW_PREFIX}/sbin:#{ENV["PATH"]}"


lib/test_bot.rb
```diff
@@ -57,7 +57,7 @@ def run!(args)
       ENV["HOMEBREW_NO_EMOJI"] = "1"
       ENV["HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK"] = "1"
       ENV["HOMEBREW_FAIL_LOG_LINES"] = "150"
-      ENV["HOMEBREW_CURL_PATH"] = "/usr/bin/curl"
+      ENV["HOMEBREW_CURL_PATH"] = "/usr/bin/curl" if ENV["HOMEBREW_CURL_PATH"].nil?
```
carlocab, line 60: We should just set this to the value we want instead of allowing it to be configured from the environment. Something like
```ruby
ENV["HOMEBREW_CURL_PATH"] = ENV["HOMEBREW_ON_DEBIAN7"].present? ? "/opt/bin/curl" : "/usr/bin/curl"
```
will probably do.

carlocab (approved):

Bo98:
Does `HOMEBREW_FORCE_BREWED_CA_CERTIFICATES` not cover this?

iMichka:
I gave `HOMEBREW_FORCE_BREWED_CA_CERTIFICATES`, but it did not work.

```
2022-02-20T10:15:55.6692632Z ##[group]Full fetch binutils --build-bottle output
2022-02-20T10:15:55.6693272Z [34m==>[0m [1mDownloading https://ftp.gnu.org/gnu/binutils/binutils-2.38.tar.xz[0m
2022-02-20T10:15:55.6693648Z curl: (60) SSL certificate problem: certificate has expired
2022-02-20T10:15:55.6693994Z More details here: http://curl.haxx.se/docs/sslcerts.html
2022-02-20T10:15:55.6694192Z 
2022-02-20T10:15:55.6694367Z curl performs SSL certificate verification by default, using a "bundle"
2022-02-20T10:15:55.6694719Z  of Certificate Authority (CA) public keys (CA certs). If the default
2022-02-20T10:15:55.6695162Z  bundle file isn't adequate, you can specify an alternate file
2022-02-20T10:15:55.6695669Z  using the --cacert option.
2022-02-20T10:15:55.6696003Z If this HTTPS server uses a certificate signed by a CA represented in
2022-02-20T10:15:55.6696568Z  the bundle, the certificate verification probably failed due to a
2022-02-20T10:15:55.6696961Z  problem with the certificate (it might be expired, or the name might
2022-02-20T10:15:55.6697295Z  not match the domain name in the URL).
2022-02-20T10:15:55.6697717Z If you'd like to turn off curl's verification of the certificate, use
2022-02-20T10:15:55.6698099Z  the -k (or --insecure) option.
2022-02-20T10:15:55.6698334Z Trying a mirror...
2022-02-20T10:15:55.6698966Z [34m==>[0m [1mDownloading https://ftpmirror.gnu.org/binutils/binutils-2.38.tar.xz[0m
2022-02-20T10:15:55.6699368Z curl: (60) SSL certificate problem: certificate has expired
2022-02-20T10:15:55.6699700Z More details here: http://curl.haxx.se/docs/sslcerts.html
2022-02-20T10:15:55.6699903Z 
2022-02-20T10:15:55.6700080Z curl performs SSL certificate verification by default, using a "bundle"
2022-02-20T10:15:55.6700445Z  of Certificate Authority (CA) public keys (CA certs). If the default
2022-02-20T10:15:55.6700891Z  bundle file isn't adequate, you can specify an alternate file
2022-02-20T10:15:55.6701239Z  using the --cacert option.
2022-02-20T10:15:55.6701541Z If this HTTPS server uses a certificate signed by a CA represented in
2022-02-20T10:15:55.6701913Z  the bundle, the certificate verification probably failed due to a
2022-02-20T10:15:55.6702608Z  problem with the certificate (it might be expired, or the name might
2022-02-20T10:15:55.6702962Z  not match the domain name in the URL).
2022-02-20T10:15:55.6703395Z If you'd like to turn off curl's verification of the certificate, use
2022-02-20T10:15:55.6703743Z  the -k (or --insecure) option.
```

I don't see `ca-certificates` being installed, I need to investigate that.

Bo98:
It worked a few months back for the Portable Ruby CI. That should be a good indication whether it still works.

iMichka:
I tested locally in the current wheezy container, and I can checkout the current binutils PR, and run either `brew fetch/brew install`, and no need to do anything special, it will just work. I don't understand why it's different when run in the homebrew-core CI. Except if test-bot is doing something fancy.

carlocab:
Not sure what exactly you mean by not having to do anything special, but installing `ca-certificates` only works automatically on macOS. We probably need to fix `install_ca_certificates_if_needed` in `lib/tests/formulae.rb` to check the environment variable.

iMichka:
What I mean is that you can run:
`docker run -it --rm homebrew/debian7`
`cd /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core/`
`git remote add pt https://github.com/pthariensflame/homebrew-core`
`git fetch pt`
`git checkout binutils-2.38-only`
`brew fetch -sv binutils OR brew install binutils -svd`

and it will just download the source file without complaining.

But doing the same with test bot will complain about broken certificates.

iMichka:
Interestingly if I run `brew test-bot binutils` inside the wheezy container on my computer, it works. But not on the CI machine.

iMichka:
I had a look at the code, `HOMEBREW_FORCE_BREWED_CA_CERTIFICATES` is set in wheezy, `ca-certificates` are pre-installed by our wheezy container. The curl shim in brew are correctly used. I am going to roll back the curl changes we made, I think the issue is elsewhere.

Bo98:
What probably happens is `--cleanup` removes the `ca-certificates` installation. We probably want to adjust https://github.com/Homebrew/homebrew-test-bot/blob/master/lib/tests/formulae.rb#L45 to be something like `return if DevelopmentTools.ca_file_handles_most_https_certificates? && !Homebrew::EnvConfig.force_brewed_ca_certificates?`

Bo98:
Or alternatively adjust `DevelopmentTools.ca_file_handles_most_https_certificates?` to return false when `ca-certificates` is needed. This is better, but trickier to do in a generic way.

iMichka:
Do you mean the `brew test-bot --only-cleanup-before` step that is run on CI? Interesting, did not think about that.

But I tried locally in the wheezy container, and  `ca-certificates` stays installed.

Bo98:
If you are running locally, I think you need to pass --cleanup for it to actually do destructive cleanup. It is automatically enabled on GitHub Actions.

iMichka:
Haaa. Ok this was surprising, https://github.com/Homebrew/homebrew-test-bot/blob/261e5e276066e1ec72776111cc60f4df0731678f/cmd/test-bot.rb#L104-L116 is maybe still doing too much magic, and we might want to be more explicit and remove some of these implicit additions. Especially that you don't see in the logs that --cleanup was passed, so it's harder to debug when you don't know what you are doing.
If you want me to cleanup stuff a little bit here, I can do it (will need to adjust some build pipelines but that's fine).

I confirmed with `brew list` in homebrew-core that the certificates were uninstalled.

I've opened a PR to force the usage/installation of ca-certificates on Debian 7: https://github.com/Homebrew/brew/pull/12907

MikeMcQuaid:
> is maybe still doing too much magic, and we might want to be more explicit and remove some of these implicit additions. Especially that you don't see in the logs that --cleanup was passed, so it's harder to debug when you don't know what you are doing.

I'd rather we leave this as-is, it's more useful to adjust this behaviour rather than have the defaults not make sense on GitHub Actions' ephemeral runners.

Definitely game for having more output that clarifies this behaviour, though.

PR #792, opened 2022-03-05
add keyword_argument to `FormulaInstaller#new`

hyuraku:
added keyword argument `build_bottle` to `FormulaInstaller.new` in order to delete `attr_writer` to be deleted which is stated in the comment https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula_installer.rb#L45


MikeMcQuaid:
Thanks @hyuraku!

PR #804, opened 2022-05-12
Fix misnamed variable

samford:
This is a follow-up to #802, where I reworked some of the livecheck tests. It was [brought to my attention](https://github.com/Homebrew/homebrew-test-bot/commit/6e5b9dbd11338e55995293df9852ece37123af3d#r73450218) that I accidentally referred to the `livecheck_info` variable as `livecheck_json` in a couple places and this is predictably causing issues. This PR resolves the issue by renaming the misnamed variables.


shivammathur:
@samford Thanks for fixing this quickly

PR #824, opened 2022-08-30
formulae: run `brew deps --tree`

MikeMcQuaid:
This is useful for debugging and will detect circular dependencies.

Fixes https://github.com/Homebrew/brew/issues/13774


carlocab (approved):
Makes sense to me, but I think we need this to fail when there is a circular dependency. I don't think it does that currently.

PR #844, opened 2022-10-20
Pass --local always when in GitHub Actions environment

dawidd6:
See discussion in: https://github.com/Homebrew/homebrew-core/pull/113506

Closes: https://github.com/Homebrew/homebrew-core/pull/113506

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -106,14 +106,12 @@ module Homebrew
   def setup_argv_and_env
     github_actions = ENV["GITHUB_ACTIONS"].present?
     if github_actions
-      ARGV << "--cleanup"
+      ARGV << "--cleanup" << "--local"
       ENV["HOMEBREW_COLOR"] = "1"
       ENV["HOMEBREW_GITHUB_ACTIONS"] = "1"
     end
 
     # TODO: remove me after a few days
     ENV["GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED"] = "1" if ENV["GITHUB_ACTIONS_HOMEBREW_MACOS_SELF_HOSTED"].present?
-
-    ARGV << "--local" if ENV["GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED"].present?
   end
 end
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -110,8 +110,5 @@ module Homebrew
       ENV["HOMEBREW_COLOR"] = "1"
       ENV["HOMEBREW_GITHUB_ACTIONS"] = "1"
     end
-
-    # TODO: remove me after a few days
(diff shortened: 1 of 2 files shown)


cmd/test-bot.rb
```diff
@@ -106,14 +106,12 @@ def test_bot
...
     github_actions = ENV["GITHUB_ACTIONS"].present?
     if github_actions
-      ARGV << "--cleanup"
+      ARGV << "--cleanup" << "--local"
       ENV["HOMEBREW_COLOR"] = "1"
       ENV["HOMEBREW_GITHUB_ACTIONS"] = "1"
     end
 
     # TODO: remove me after a few days
     ENV["GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED"] = "1" if ENV["GITHUB_ACTIONS_HOMEBREW_MACOS_SELF_HOSTED"].present?
```
MikeMcQuaid, line 115:
```suggestion
```

MikeMcQuaid (approved):
Fine with me! Be ready to rollback if we need to. Would be nice to remove the TODO now too.

MikeMcQuaid (approved):

dawidd6:
MacOS CI fails, can't figure out why.

MikeMcQuaid:
@dawidd6 it's trying to write logs somewhere it doesn't have permission to: 
```
    Errno::EPERM: Operation not permitted @ rb_sysopen - /usr/local/Homebrew/Library/Taps/homebrew/homebrew-test-bot/logs/testbottest/post_install.01.testbottest
```

MikeMcQuaid:
(it's also a bad idea to be writing logs to e.g. inside `/usr/local` anyway)

MikeMcQuaid:
@dawidd6 We shouldn't be putting bottle files there. They should be in e.g. $HOME

Bo98:
I'm a bit concerned with the compatibility of this change. Particularly since most taps will be using `brew test-bot` (we configure it in `brew tap-new`).

If this is requiring changes to workflows then I don't think it's worth it and could be passed in Homebrew/core like we do with many other things like `--junit`. Having to change the workflow here is already an indicator.

Bo98:
Honestly the main reason `--local` is forced on this side is because of the non-ephemeral nature of our self-hosted runners, so that conditional can probably go when fully ephemeral.

Beyond that it's only useful because of the specific way we upload artifacts on Homebrew/core, so probably makes sense controlled there.

dawidd6:
Yea I agree, we should fix this on the core tap level rather than here. Imma close this PR and go back to the previous one there.

MikeMcQuaid:
@dawidd6 Could you reopen just to remove the TODO? Thanks 🙇🏻

dawidd6:
https://github.com/Homebrew/homebrew-test-bot/pull/845

PR #850, opened 2022-11-04
tests/formulae: use --no-all-checks

Bo98:
See https://github.com/Homebrew/brew/pull/14095

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -176,7 +176,7 @@ module Homebrew
         @bottle_json_filename =
           @bottle_filename.gsub(/\.(\d+\.)?tar\.gz$/, ".json")
         bottle_merge_args =
-          ["--merge", "--write", "--no-commit", @bottle_json_filename]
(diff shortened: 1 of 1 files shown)


lib/tests/formulae.rb
```diff
@@ -176,7 +176,7 @@ def bottle_reinstall_formula(formula, new_formula, args:)
         @bottle_json_filename =
           @bottle_filename.gsub(/\.(\d+\.)?tar\.gz$/, ".json")
         bottle_merge_args =
-          ["--merge", "--write", "--no-commit", @bottle_json_filename]
+          ["--merge", "--write", "--no-commit", "--no-all-checks", @bottle_json_filename]
```
MikeMcQuaid, line 179: @Bo98 Am I misreading or has this unconditionally turned off `:all` bottles being created in all circumstances? If so, how are we getting back to these bottles being created again?

MikeMcQuaid: Ignore me, reread https://github.com/Homebrew/brew/pull/14095. Might be good to clarify my question in a comment here just to avoid confusion?

PR #853, opened 2022-11-24
tests/formulae: make dependency tree failures an annotation

Bo98:
Not tested, but this should hopefully prevent reoccurrences of https://github.com/Homebrew/homebrew-core/pull/115951.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -296,7 +296,7 @@ module Homebrew
           return
         end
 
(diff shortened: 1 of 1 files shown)


Bo98:
macOS failure is unrelated but can't guarantee I'll be able to look into it before the weekend.

carlocab (approved):

MikeMcQuaid (approved):

PR #873, opened 2023-02-17
Fix errors when Homebrew/core is untapped

Bo98:
Needed for Homebrew/actions#332.

Allowing at least the bare minimum of this to work without Homebrew/core tapped seems sensible.

This fix is for non-core CI primarily as we should be tapped still for core CI.

### DIFF
--- a/lib/tests/cleanup_before.rb
+++ b/lib/tests/cleanup_before.rb
@@ -6,15 +6,8 @@ module Homebrew
       def run!(args:)
         test_header(:CleanupBefore)
 
-        if tap.to_s != CoreTap.instance.name
-          core_path = CoreTap.instance.path
-          if core_path.exist?
-            reset_if_needed(core_path.to_s)
-          else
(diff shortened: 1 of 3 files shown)


lib/tests/cleanup_before.rb
```diff
@@ -6,15 +6,8 @@ class CleanupBefore < TestCleanup
...
         test_header(:CleanupBefore)
 
-        if tap.to_s != CoreTap.instance.name
-          core_path = CoreTap.instance.path
-          if core_path.exist?
-            reset_if_needed(core_path.to_s)
-          else
-            test git, "clone",
-                 CoreTap.instance.default_remote,
-                 core_path.to_s
```
Bo98, line 16: It would always error before getting to this point so this was dead code.

Bo98:
Merging to unblock CI.

MikeMcQuaid:
Nice work here @Bo98!

PR #874, opened 2023-02-17
test_cleanup: handle case untapped Homebrew/core edge case

Bo98:
To cover the cases where you use `--only-cleanup*` outside of a tap (like we do in Homebrew/brew).

### DIFF
--- a/lib/test_cleanup.rb
+++ b/lib/test_cleanup.rb
@@ -46,9 +46,11 @@ module Homebrew
     def cleanup_shared
       FileUtils.chmod_R("u+X", HOMEBREW_CELLAR, force: true)
 
-      cleanup_git_meta(repository)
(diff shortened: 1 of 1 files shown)


lib/test_cleanup.rb
```diff
@@ -46,9 +46,11 @@ def delete_or_move(paths)
     def cleanup_shared
       FileUtils.chmod_R("u+X", HOMEBREW_CELLAR, force: true)
 
-      cleanup_git_meta(repository)
-      clean_if_needed(repository)
-      prune_if_needed(repository)
+      if repository.exist?
+        cleanup_git_meta(repository)
```
MikeMcQuaid, line 50: IMO these should be handled in the methods themselves but: fine with me, no need to change unless you agree. Thanks for the fix!

PR #879, opened 2023-03-02
lib/tests/formulae: pass `--formula` to `brew livecheck`

nandahkrishna:
This is to avoid the failure seen in https://github.com/Homebrew/homebrew-core/pull/124651. We should explicitly pass `--formula` so that the `JSON.parse` doesn't error out when there's a cask of the same name:
```
Error: unexpected token at 'Warning: Treating puzzles as a formula. For the cask, use homebrew/cask/puzzles
[
  {
    "formula": "puzzles",
    "version": {
      "current": "20230302",
      "latest": "20230302",
      "outdated": false,
      "newer_than_upstream": false
    }
  }
]
'
```

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -225,7 +225,7 @@ module Homebrew
         return unless formula.livecheckable?
         return if formula.livecheck.skip?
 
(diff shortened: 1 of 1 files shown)


chenrui333 (approved):

MikeMcQuaid:
Good catch, thanks @nandahkrishna!

PR #898, opened 2023-04-04
Properly stub resolved bottle filename.

reitermarkus:
Previously, this assumed that the bottle filename was resolved by falling back to the path in the URL, e.g. `0.1.tgz` if there was a server error.

Fixes the CI error in https://github.com/Homebrew/brew/pull/15117, which makes resolving the file name a hard error in case the server returns e.g. 404.


carlocab (approved):

MikeMcQuaid:
Makes sense, thanks @reitermarkus!

PR #901, opened 2023-04-07
Revert "Merge pull request #899 from Homebrew/formula_detect_cask"

muzimuzhi:
This PR reverts PR #899 which fixed issue #897.

The cause of #897 was later identified as coming from `Homebrew/brew` itself, and has been fixed in
- https://github.com/Homebrew/brew/pull/15150

Thus in `homebrew-test-bot` here #899 can be safely reverted.


MikeMcQuaid:
Thanks for cleaning up, great work again @muzimuzhi!

PR #921, opened 2023-05-12
tests/tap_syntax: safer API generation.

MikeMcQuaid:
- Make checks a bit stricter
- Only remove directory if we created it


Bo98:
Thanks!

PR #951, opened 2023-06-22
tests: improve no-API switching for named args

Bo98:
Everything in a CI environment worked ok. This is just small tweaks so that a `brew test-bot hello` on the command line switches to no-API mode.


MikeMcQuaid (approved):
Makes sense, thanks!

PR #966, opened 2023-09-02
tests: speed up non-push/merge group cases.

MikeMcQuaid:
- only run `readall` or `audit` on the whole tap for `push` and `merge_group` runs.
- always run `brew style` on the formula in case `tap_syntax` isn't run or there's macOS/Linux differences as it's really fast.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -528,6 +528,7 @@ module Homebrew
 
         livecheck(formula) if !args.skip_livecheck? && !skip_online_checks
 
+        test "brew", "style", formula_name
         test "brew", "audit", *audit_args unless formula.deprecated?
         unless install_step_passed
           if ignore_failures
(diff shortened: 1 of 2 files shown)


p-linnane (approved):

chenrui333 (approved):

Bo98 (approved):
This is likely going to be annoying for recursive dependency checks, or CI-syntax-only PRs, where it will only fail when merging.

Though I'm ok with this for now until we do something like `brew audit --readall`.

MikeMcQuaid:
Thanks for review, all!

> This is likely going to be annoying for recursive dependency checks, or CI-syntax-only PRs, where it will only fail when merging.

Is that less annoying than having an extra >5m on every PR? It's all trade-offs here.



> Though I'm ok with this for now until we do something like `brew audit --readall`.

I'm not sure what this would be and how it would work?

PR #968, opened 2023-09-15
Improve dependent test ordering

Bo98:
This restores changes made in https://github.com/Homebrew/homebrew-test-bot/commit/fab38f1bfad42221f7af06a58ca72b62228f5eb6 that were lost in https://github.com/Homebrew/homebrew-test-bot/commit/cbe87b4b5a7ad44b70ea34d593c797b660db5f0d.

In summary, this PR:
* Changes the order of formulae we run dependent testing on to be the same order as the formulae were built. This means it's roughly topological now so the postinstalls will now be run in the correct order should one formula we're testing the dependents of is a dependency of another such formula.
* Dependents of multiple formulae will now be deferred as much as possible rather than eagerly tested when they're first encountered. This again ensures we're only testing things after all the relevant postinstalls have happened.

Should fix issues seen in https://github.com/Homebrew/homebrew-core/pull/141617.

### DIFF
--- a/lib/test_formulae.rb
+++ b/lib/test_formulae.rb
@@ -362,6 +362,46 @@ module Homebrew
 
         test "brew", "uninstall", "--force", "--ignore-dependencies", *uninstallable_formulae
       end
+
+      def sorted_formulae
+        changed_formulae_dependents = {}
+
+        @testing_formulae.each do |formula|
+          begin
+            formula_dependencies =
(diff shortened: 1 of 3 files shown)


MikeMcQuaid (approved):
Looks good, thanks @Bo98!

iMichka (approved):
Let's try this out. Thanks!

PR #969, opened 2023-09-17
Add os/arch options to readall command

apainintheneck:
This is a follow-up to changes that added os/arch combinations a few months ago. They were made the default temporarily until the references to `brew readall` in the other Homebrew repos were updated as well.

- https://github.com/Homebrew/brew/pull/15470
- https://github.com/Homebrew/brew/pull/15937#discussion_r1313889254


reitermarkus (approved):

MikeMcQuaid (approved):
Makes sense, thanks @apainintheneck!

PR #973, opened 2023-09-28
tests/tap_syntax: simplify and speed up

Bo98:
After perf and coverage improvments, we can simplify this for an overall speed boost.

I've also explicitly added a `--except=installed` here. It's not needed for most of our CI, but might be on `macos-latest`. The result of tap_syntax shouldn't be dependent on what formulae happen to be installed - these checks are best suited (and are run) on a per-formula basis in formula tests. We do not guarantee the speed of installed-formula audits, as it largely varies on how big formulae are and all our perf improvements haven't considered installed-formula audits.

### DIFF
--- a/lib/tests/tap_syntax.rb
+++ b/lib/tests/tap_syntax.rb
@@ -9,32 +9,10 @@ module Homebrew
 
         test "brew", "style", tap.name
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Makes sense to me, thanks again @Bo98!

PR #975, opened 2023-10-14
Update rspec Ruby, Gemfile, actions/cache, etc.

Bo98:
We're moving away from 2.6 and the rspecs tests here don't _need_ to be running 2.6 still, particularly when we still have `brew test-bot` commands being run in the workflow which will be use Homebrew's Ruby. It might also give us an early warning of Ruby 3.1 issues, though the rspec test coverage here is extremely minimal.

Also update `actions/cache` to v3 since we support that now in setup-homebrew.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,8 +6,8 @@ updates:
     interval: daily
   open-pull-requests-limit: 10
   allow:
-  - dependency-type: direct
-  - dependency-type: indirect
+    - dependency-type: all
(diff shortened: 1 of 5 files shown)


Bo98:
Ha, turns out it's been testing `master` - not the PR branch!

MikeMcQuaid (approved):
Works for me, thanks @Bo98!

SMillerDev (approved):

PR #977, opened 2023-10-23
tests/formulae_detect: check for formulae with version or bottle changes

Bo98:
Idea here is to only run fetch checks in the merge queue for formulae where the `pkg_version` or bottle block have changed. This exempts e.g. `brew style` changes from having to go through bottle fetch tests.

Requires workflow changes to take effect, but the logic can be monitored to see if it's working before that since we print the output here.

Requires Homebrew/brew#16130.

### DIFF
--- a/lib/tests/formulae_detect.rb
+++ b/lib/tests/formulae_detect.rb
@@ -11,6 +11,7 @@ module Homebrew
         @argument = argument
         @added_formulae = []
         @deleted_formulae = []
+        @formulae_to_fetch = []
       end
 
       def run!(args:)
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Makes sense to me! I assume this isn't slow but may be worth checking before/after merge.

Bo98:
> I assume this isn't slow

Will see in practice, but seems to be maybe ~0.1s per formula. Adds up to one and a half minutes for 1000 formulae, but you would have several hours of fetching without it so is an overall gain.

I've modified this to scope the checks to only apply on `merge_group` to make sure this doesn't regress pull request CI runtime.

MikeMcQuaid:
> Will see in practice, but seems to be maybe ~0.1s per formula. Adds up to ~1.5 minutes for 1000 formulae, but you would have several hours of fetching without it so is an overall gain.

HUGE gain. Well done @Bo98.

PR #978, opened 2023-10-28
test_runner: scope JUnit output to only formula steps

Bo98:
REXML will not be available with Ruby 3 and JUnit output currently needs it. However, we only really need JUnit output when doing the formula steps (because we filter for `audit` and `test`), and formula steps assume that `brew test-bot --only-setup` has already been run with the `brew install-bundler-gems --groups=...` it internally does, so simply making sure it only is executed when running those steps should address the issue.


MikeMcQuaid (approved):
Makes sense, thanks again @Bo98!

PR #984, opened 2023-11-21
Gate `MacOS` calls behind `OS.mac?` conditionals

Rylan12:
See https://github.com/Homebrew/brew/pull/16224

This PR makes it so that all references to the `MacOS` module are gates behind `OS.mac?` calls to allow the `MacOS` module to be deprecated on Linux.

My only outstanding question is whether this reference needs to be changed:

https://github.com/Homebrew/homebrew-test-bot/blob/9da1d1f7f4de703c517c22dbdc237725ed4cd5ac/spec/stub/os.rb#L3-L9

This is in the test directory so it shouldn't affect functionality, only CI in this repo.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -242,7 +242,7 @@ module Homebrew
 
         # This is needed where sparse files may be handled (bsdtar >=3.0).
         # We use gnu-tar with sparse files disabled when --only-json-tab is passed.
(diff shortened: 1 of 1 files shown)


Bo98:
> My only outstanding question is whether this reference needs to be changed:

It doesn't need to be changed, no. You could make it conditional to Darwin to force it to error on Linux if encountered, but that doesn't matter much.

The rspecs in this repo is very poor anyway atm and hardly do anything.

Bo98 (approved):

MikeMcQuaid:
Thanks @Rylan12 for the PR and @Bo98 for the review and merge!

PR #996, opened 2024-01-03
Fix `ignore_failures` usage with `--test-default-formula`

MikeMcQuaid:
Previously we were ignoring a bunch of failures in these cases which makes this CI check pretty useless (e.g. see https://github.com/Homebrew/brew/pull/16423).

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -540,7 +540,7 @@ module Homebrew
           # they can be unavoidable but we still want to know about them.
           test "brew", "linkage", "--cached", "--test", "--strict",
                named_args:      formula_name,
-               ignore_failures: !args.test_default_formula? && true
+               ignore_failures: !args.test_default_formula?
         end
 
         test "brew", "linkage", "--cached", formula_name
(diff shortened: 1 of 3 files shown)


lib/tests/formulae.rb
```diff
@@ -540,7 +540,7 @@ def formula!(formula_name, args:)
           # they can be unavoidable but we still want to know about them.
           test "brew", "linkage", "--cached", "--test", "--strict",
                named_args:      formula_name,
-               ignore_failures: true
+               ignore_failures: !args.test_default_formula? && true
```
Bo98, line 543: `&& true` is redundant

Bo98 (approved):
If this is every one, I wonder if we should just do this in `test.rb` to prevent future occurrences.

carlocab (approved):

lib/tests/formulae.rb
MikeMcQuaid:
```suggestion
               ignore_failures: !args.test_default_formula?
```

PR #1033, opened 2024-04-05
test-bot: allow skipping repository audit

p-linnane:
We often need to skip repository checks (repo age, notability, etc.) for various reasons. This PR allows us to skip those checks with a switch, so that we can reduce the number of ❌ PR's we are merging in homebrew-core.

### DIFF
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -427,10 +427,11 @@ module Homebrew
         audit_args = [formula_name]
         audit_args << "--online" unless skip_online_checks
         audit_args << "--except=unconfirmed_checksum_change" if args.skip_checksum_only_audit?
+        audit_args << "--except=github_repository,gitlab_repository,bitbucket_repository" if args.skip_repository_audit?
         audit_args << "--except=stable_version" if args.skip_stable_version_audit?
         audit_args << "--except=revision" if args.skip_revision_audit?
         if new_formula
-          audit_args << "--new" unless args.skip_repository_audit?
+          audit_args << "--new"
         else
           audit_args << "--git" << "--skip-style"
         end
--- a/lib/tests/formulae.rb
+++ b/lib/tests/formulae.rb
@@ -428,8 +428,8 @@ module Homebrew
 
         audit_args = [formula_name]
         audit_args << "--online" unless skip_online_checks
-        audit_args << "--except=unconfirmed_checksum_change" if args.skip_checksum_only_audit?
         audit_args << "--except=#{repo_audits}" if args.skip_repository_audit?
+        audit_args << "--except=unconfirmed_checksum_change" if args.skip_checksum_only_audit?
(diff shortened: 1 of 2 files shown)


lib/tests/formulae.rb
```diff
@@ -442,7 +442,7 @@ def formula!(formula_name, args:)
         audit_args << "--except=stable_version" if args.skip_stable_version_audit?
         audit_args << "--except=revision" if args.skip_revision_audit?
         if new_formula
-          audit_args << "--new"
+          audit_args << "--new" unless args.skip_repository_audit?
```
ZhongRuoyu, line 445: We probably don't want to skip `--new` entirely because `--new` implies `--strict` and `--online` which include a bunch of other audits too (e.g., "Formulae in homebrew/core must specify a license.").

To really skip the repository audits, I suggest adding `--except=github_repository,gitlab_repository,bitbucket_repository` to the audit args instead. They should cover this category of new-formula audit failures we usually see in `homebrew/core`. See:

- https://github.com/Homebrew/brew/blob/567386a13ead16e510d228e93b0d43c1677e9d0b/Library/Homebrew/formula_auditor.rb#L629-L658
- https://github.com/Homebrew/brew/blob/567386a13ead16e510d228e93b0d43c1677e9d0b/Library/Homebrew/utils/shared_audits.rb#L105-L166

p-linnane: Good call. I only saw the ones in `shared_audits` on first look. I've pushed the change.

lib/tests/formulae.rb
```diff
@@ -424,9 +424,12 @@ def formula!(formula_name, args:)
...
         fetch_args << "--force" if args.cleanup?
 
+        repo_audits = "github_repository,gitlab_repository,bitbucket_repository"
+
         audit_args = [formula_name]
         audit_args << "--online" unless skip_online_checks
         audit_args << "--except=unconfirmed_checksum_change" if args.skip_checksum_only_audit?
+        audit_args << "--except=#{repo_audits}" if args.skip_repository_audit?
         audit_args << "--except=stable_version" if args.skip_stable_version_audit?
         audit_args << "--except=revision" if args.skip_revision_audit?
```
ZhongRuoyu, line 434: Just one nit and I'm good with this: let's reorder the `--except` args so that they appear in the same order as the corresponding CLI flags.

```suggestion
        audit_args << "--except=#{repo_audits}" if args.skip_repository_audit?
        audit_args << "--except=unconfirmed_checksum_change" if args.skip_checksum_only_audit?
        audit_args << "--except=stable_version" if args.skip_stable_version_audit?
        audit_args << "--except=revision" if args.skip_revision_audit?
```

ZhongRuoyu (approved):
Thanks!

chenrui333 (approved):
lgtm 👍

PR #1036, opened 2024-04-13
Add stable option to restrict tap syntax checks

cho-m:
Related to https://github.com/Homebrew/homebrew-core/pull/166913

Wasn't sure if we wanted a mutually exclusive arg (and optionally a separate Test) or just a dependent arg. Went with the simplest change which is `--stable` option.

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -71,6 +71,9 @@ module Homebrew
              description: "Only run the local system setup check step."
       switch "--only-tap-syntax",
              description: "Only run the tap syntax check step."
+      switch "--stable",
+             depends_on:  "--only-tap-syntax",
+             description: "Only run the tap syntax checks needed on stable brew."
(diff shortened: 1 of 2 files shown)


MikeMcQuaid (approved):
Makes sense, thanks again @cho-m!

cmd/test-bot.rb
```diff
@@ -71,6 +71,9 @@ def test_bot_args
              description: "Only run the local system setup check step."
       switch "--only-tap-syntax",
              description: "Only run the tap syntax check step."
+      switch "--stable",
+             depends_on:  "--only-tap-syntax",
+             description: "Only run the tap syntax checks needed on stable brew."
```
MikeMcQuaid, line 76: Whoops, only thought more after merge: I think this is maybe a bit unclear for third-party taps as to what it's doing? What about making this `--skip-audit` instead?

PR #1041, opened 2024-04-23
test-bot: allow skipping audit `--new` or `--strict` flags.

MikeMcQuaid:
- `--skip-new` skips the `--new` flag and uses `--strict` instead for new formulae.
- `--skip-strict` also skips the `--strict` flag for new formulae when `--skip-new` is used.
- remove `--skip-repository-audit` which is covered now by `--skip-new`.
- move other audit args that only apply for non-new formulae into a better place.


p-linnane (approved):
Thanks!

PR #1057, opened 2024-05-09
tests/setup: install more bundler gems.

MikeMcQuaid:
These are all needed by various stages of installing formulae.


chenrui333:
Thanks @MikeMcQuaid!

relates to https://github.com/Homebrew/homebrew-core/pull/171273

PR #1062, opened 2024-05-14
tests.yml: adjust `rm` command

fxcoudert:
If a bottle starts with a `-`, this could be interpreted as an option. Use `--` to mark end of options.


carlocab (approved):

MikeMcQuaid:
Makes sense, thanks @fxcoudert!

PR #1095, opened 2024-06-06
cmd/test-bot: migrate to AbstractCommand

MikeMcQuaid:
This is a nicer format and will get analytics.

### DIFF
--- a/cmd/test-bot.rb
+++ b/cmd/test-bot.rb
@@ -1,132 +1,124 @@
 # frozen_string_literal: true
 
-require "cli/parser"
+require "abstract_command"
+require_relative "../lib/test_bot"
+
 module Homebrew
-  module_function
+  module Cmd
+    class TestBotCmd < AbstractCommand
+      cmd_args do
+        usage_banner <<~EOS
+          `test-bot` [<options>] [<formula>]
 
-  def test_bot_args
-    Homebrew::CLI::Parser.new do
-      usage_banner <<~EOS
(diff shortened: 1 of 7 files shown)


cmd/test-bot.rb
```diff
@@ -1,132 +1,131 @@
...
-
-  def setup_argv_and_env
-    return if ENV["GITHUB_ACTIONS"].blank? && ENV["GITHUB_ACTIONS_HOMEBREW_SELF_HOSTED"].blank?
-
-    ARGV << "--cleanup" << "--local"
-    ENV["HOMEBREW_COLOR"] = "1"
-    ENV["HOMEBREW_GITHUB_ACTIONS"] = "1"
+        ARGV << "--cleanup" << "--local"
+        ENV["HOMEBREW_COLOR"] = "1"
+        ENV["HOMEBREW_GITHUB_ACTIONS"] = "1"
```
MikeMcQuaid, line 127: @dduugg This approach doesn't quite work. Any thoughts on how best to handle it?

p-linnane (approved):

cmd/test-bot.rb
dduugg: I see this went in, I'm not sure what your issue was. I know working with `ARGV` can be a pain, the `irb` command has some workarounds that might be helpful to reference.

MikeMcQuaid: @dduugg Thanks for replying! Yeh, @Bo98 suggested just not doing this which was easier than expected.

PR #1142, opened 2024-07-14
test_formulae: only allow fork build cache usage for BrewTestBot forks.

MikeMcQuaid:
Maintainers are now encouraged to branch directly on homebrew-core.


Bo98:
BrewTestBot doesn't even have any forks anymore

MikeMcQuaid:
@Bo98 Yes but, if that changes in future: we should allow it. I think we should moving back to forks there so it doesn't need to have write permissions on homebrew-core at all.

Bo98 (approved):

MikeMcQuaid:
Thanks @Bo98!

PR #1184, opened 2024-08-22
Pass `git` into `tap_syntax` tests

issyl0:
- So that `git -C /path/to/tap grep ...` works correctly to determine if a tap should be typechecked.
- Thanks for your patience with my various test PRs, here's it working: https://github.com/Homebrew/homebrew-command-not-found/actions/runs/10516250377/job/29138543034.


carlocab (approved):

MikeMcQuaid:
Thanks for figuring this out @issyl0!

PR #1185, opened 2024-08-22
Skip `brew typecheck` if running on stable

issyl0:
- Since `brew typecheck <tap>` is not in a `brew` release yet, we need to skip this for stable `tap_syntax` checks otherwise: https://github.com/Homebrew/homebrew-core/actions/runs/10516750573/job/29139564452?pr=182076


issyl0:
> Style/NegatedIf: Favor unless over if for negative conditions.

Keeping up with when `unless` is preferred or not is really difficult.

carlocab (approved):

PR #1192, opened 2024-08-27
Update usage of `Tap#formula_file?`

apainintheneck:
Currently this method takes either a String or a Pathname which is fine since we were passing in a Pathname. I'm planning on changing that behavior in new PR so that it will only accept the String class going forward so I'm updating this to use the String class here. This should be completely backwards compatible.

- Relevant PR: https://github.com/Homebrew/brew/pull/18167

I'm not too familiar with this repo so I'm not entirely sure how to test this change.


carlocab (approved):

MikeMcQuaid:
Looks good, thanks!

PR #1238, opened 2024-09-26
Don't verify attestations for 3rd-party taps

scpeters:
The CI machines for 3rd-party taps might not have a gh token and may fail to verify attestations, which is the case for us in https://github.com/osrf/homebrew-simulation/issues/2815.

Follow-up to https://github.com/Homebrew/homebrew-test-bot/pull/1230.


carlocab (approved):
Thanks!

PR #1257, opened 2024-10-05
Replace removable constants with overridable methods

dduugg:
Will resolve failing checks in https://github.com/Homebrew/brew/pull/18517

These PRs need to merged together to work. (I can also update the linked PR to support both constants and methods as an intermediate step.)

### DIFF
--- a/lib/test_cleanup.rbi
+++ b/lib/test_cleanup.rbi
@@ -1,7 +1,0 @@
-# typed: strict
-
-module Homebrew
-  class TestCleanup
-    include Kernel
-  end
-end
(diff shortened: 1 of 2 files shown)


lib/test_cleanup.rbi
```diff
@@ -1,7 +0,0 @@
-# typed: strict
-
-module Homebrew
-  class TestCleanup
-    include Kernel
```
dduugg, line 5: It shouldn't be necessary `include Kernel` in a class (`Class` is a subclass of `Object`, which includes `Kernel`… 🤕)

MikeMcQuaid:
> (I can also update the linked PR to support both constants and methods as an intermediate step.)

yes please to intermediate step

MikeMcQuaid (approved):
Looks good when 🟢!

PR #1306, opened 2024-11-18
tests/formulae: add missing `--ignore-dependencies` option

ZhongRuoyu:
One more fix like https://github.com/Homebrew/homebrew-test-bot/pull/1305.

Fixes failure seen at https://github.com/Homebrew/homebrew-core/actions/runs/11889979173/job/33128643880?pr=198048.

PR #1307, opened 2024-11-20
tests/formulae_dependents: another `--ignore-dependencies` fix

ZhongRuoyu:
One more (and hopefully last) fix like #1305 and #1306. When we test dependents of multiple formulae, a dependent of one formula being tested can happen to be a dependency of another formula being tested. In this case, we need the `--ignore-dependencies` option to force removal of the dependent/dependency formula. When testing the dependents of the removed formula, `brew install --only-dependencies` ensures that it is reinstalled.

Fixes failure seen at https://github.com/Homebrew/homebrew-core/actions/runs/11923246241/job/33242018653?pr=198240.

### DIFF
--- a/lib/tests/formulae_dependents.rb
+++ b/lib/tests/formulae_dependents.rb
@@ -326,7 +326,7 @@ module Homebrew
           test_step = steps.last
         end
 
-        test "brew", "uninstall", "--force", dependent.full_name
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @ZhongRuoyu!

PR #1353, opened 2025-01-04
formulae: use livecheck --autobump option

samford:
We recently modified the behavior of `brew livecheck` to skip packages in the tap's autobump list, so this change is causing autobumped formulae to fail the `brew livecheck` part of CI testing. This adds the `--autobump` option to the `brew livecheck` call, which will ensure that livecheck doesn't skip autobumped formulae.


chenrui333 (approved):

chenrui333:
also relates to https://github.com/Homebrew/brew/pull/18984

PR #1406, opened 2025-02-19
Add HOMEBREW_TEST_BOT environment variable

MikeMcQuaid:
This can be used in Homebrew/brew to detect if we're running inside of `brew test-bot`.

See https://github.com/Homebrew/brew/pull/19326#issuecomment-2667494086 for an example usage.


p-linnane (approved):
Makes sense to me! Thanks!

PR #1558, opened 2025-09-14
tests/setup: run `brew doctor` with `--debug`

ZhongRuoyu:
To help us catch a problem that causes CI jobs to hang.

### DIFF
--- a/lib/tests/setup.rb
+++ b/lib/tests/setup.rb
@@ -12,7 +12,11 @@ module Homebrew
         # Always output `brew config` output even when it doesn't fail.
         test "brew", "config", verbose: true
 
-        test "brew", "doctor", "--debug", verbose: true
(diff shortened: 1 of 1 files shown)


lib/tests/setup.rb
```diff
@@ -12,7 +12,7 @@ def run!(args:)
         # Always output `brew config` output even when it doesn't fail.
         test "brew", "config", verbose: true
 
-        test "brew", "doctor"
+        test "brew", "doctor", "--debug", verbose: true
```
MikeMcQuaid, line 15:
```suggestion
        if ENV["HOMEBREW_TEST_BOT_VERBOSE_DOCTOR"]
          test "brew", "doctor", "--debug", verbose: true
        else
          test "brew", "doctor"
        end
```

